PR #20, opened 2023-11-10
build: add postgresql db config (with sqlalchemy)

raphodn:
### What

- install the `psycopg2-binary` & `sqlalchemy` dependencies. Remove `peewee`
- update the .env & app config to pass the postgres db config parameters
- connect to the db on app startup
- update documentation

TODO
- docker config ?

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,3 +15,10 @@ SENTRY_DNS=
 
 # Log level to use, DEBUG by default in dev
 LOG_LEVEL=DEBUG
+
+# Postgres database
+POSTGRES_DB_NAME=open_prices
+POSTGRES_USER=open_prices_team
+POSTGRES_PASSWORD=password
+POSTGRES_HOST=localhost
+POSTGRES_PORT=5432
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -3,7 +3,7 @@
 ## Prerequisites
 
 - Python 3.10 (lower version may be OK, but untested)
-- pip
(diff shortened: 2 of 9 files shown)


.env.example
```diff
@@ -15,3 +15,10 @@ SENTRY_DNS=
 
 # Log level to use, DEBUG by default in dev
```
raphael0202, line 16: Locally we should have a ".env" file ready for local development, so you can rename this file ".env"

.env.example
```diff
@@ -15,3 +15,10 @@ SENTRY_DNS=
...
 LOG_LEVEL=DEBUG
+
+# Postgres database
+POSTGRES_DB_NAME=open_prices
```
raphael0202, line 20: We can keep it simple locally (postgres/postgres/postgres for DB name, user and password), as we use Docker usually and it's for local dev only

INSTALL.md
```diff
@@ -22,6 +22,12 @@ source venv/bin/activate
 # install
 pip install -r requirements.txt
 
+# create a PostgreSQL database
+psql -c "CREATE USER open_prices_team WITH PASSWORD 'password'"
```
raphael0202, line 26: You can skip this if you keep the defaults

raphael0202 (approved):

.env.example
raphodn: ok ! 
- i deleted `.env.example`
- remove `.env` from the gitignore
- simplified the install doc

PR #22, opened 2023-11-12
feat: authentication workflow and store user token

raphodn:
### What

- new POST endpoint `/auth`
- check that users exist via https://world.openfoodfacts.org/cgi/auth.pl
- on success, generate a token and store in the the `User` db
- return the token

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -1,4 +1,4 @@
-import time
+import asyncio
 import uuid
 from pathlib import Path
 from typing import Annotated
@@ -103,7 +103,7 @@ async def authentication(form_data: Annotated[OAuth2PasswordRequestForm, Depends
         crud.create_user(db, user=user)  # type: ignore
         return {"access_token": token, "token_type": "bearer"}
     elif r.status_code == 403:
-        time.sleep(2)   # prevents brute-force
+        await asyncio.sleep(2)   # prevents brute-force
         raise HTTPException(
             status_code=status.HTTP_401_UNAUTHORIZED,
             detail="Invalid authentication credentials",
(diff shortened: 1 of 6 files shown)


.env.example
```diff
@@ -6,6 +6,7 @@ COMPOSE_PATH_SEPARATOR=;
 COMPOSE_FILE=docker-compose.yml;docker/dev.yml
 
 API_PORT=127.0.0.1:8000
+OAUTH2_SERVER_URL=https://world.openfoodfacts.org/cgi/auth.pl
```
raphael0202, line 9: We will deploy it in staging as well (https://world.openfoodfacts.net), so the URL need not to be hardcoded. You can use openfoodfacts SDK URLBuilder (https://github.com/openfoodfacts/openfoodfacts-python/blob/develop/openfoodfacts/utils.py#L23) to build OFF URLs this easily.

app/api.py
```diff
@@ -53,6 +78,40 @@ def main_page(request: Request):
...
+
+    data = {"user_id": form_data.username, "password": form_data.password}
+    r = requests.post(settings.oauth2_server_url, data=data)  # type: ignore
+    if r.status_code == 200:
+        token = await create_token(form_data.username)
+        user: UserBase = {"user_id": form_data.username, "token": token}  # type: ignore
+        crud.create_user(db, user=user)  # type: ignore
+        return {"access_token": token, "token_type": "bearer"}
+    elif r.status_code == 403:
+        time.sleep(2)   # prevents brute-force
```
raphael0202, line 106: you should use `asyncio.sleep` (https://docs.python.org/3/library/asyncio-task.html#asyncio.sleep) otherwise you're blocking your only thread

raphael0202 (approved):

.env.example
raphodn: if it's in the .env then it's not hardcoded, right ? :sweat_smile: 
I'm not sure how to do this so I'll do it in a seperate PR :)

app/api.py
raphodn: ok thanks, done :heavy_check_mark:

PR #23, opened 2023-11-12
feat: Price model

raphodn:
### What

- install `sqlalchemy-utils` & `Babel`
- create a `User` model with sqlalchemy (basic fields: product_code, price, location, date, owner)
- make the migration with `alembic`


app/models.py
```diff
@@ -1,15 +1,45 @@
...
+    product_code = Column(String, index=True)
+
+    price = Column(Numeric(precision=10, scale=2))
+    currency = Column(CurrencyType)
+
+    location_osm_id = Column(BigInteger, index=True)
+    location_osm_type = Column(ChoiceType(PriceLocationOSMType))
+
+    date = Column(Date)
+    owner = Column(String)
```
raphael0202, line 41: what's the owner? Is it a foreign key to the User?

raphael0202 (approved):

app/models.py
raphodn: no it's just a string that store's the OFF user_id
I'd rather not make foreign keys as it's metadata

raphodn: as discussed on Slack, go for "string"

PR #25, opened 2023-11-13
build: add tests

raphodn:
### What

- install `pytest`, `pytest-cov` & `httpx`
- write basic tests on routes


pyproject.toml
```diff
@@ -23,6 +23,9 @@ sqlalchemy-utils = "~0.41.1"
 uvicorn = "~0.23.2"
 fastapi-pagination = "^0.12.12"
 fastapi-filter = "^1.0.0"
+pytest = "^7.4.3"
```
raphael0202, line 26: All dev dependencies must be added to the "dev" group (below): `poetry add --group=dev pytest`
And pytest is already included :)

tests/test_api.py
```diff
@@ -0,0 +1,140 @@
...
+SQLALCHEMY_DATABASE_URL = "sqlite://"
+
+engine = create_engine(
+    SQLALCHEMY_DATABASE_URL,
+    connect_args={"check_same_thread": False},
+    poolclass=StaticPool,
+)
+TestingSessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
+
+Base.metadata.create_all(bind=engine)
```
raphael0202, line 24: Isn't it better to do DB initialization in a global (autouse=True, scope=session) fixture? So that we can take care of removing it after

app/api.py
```diff
@@ -162,7 +161,7 @@ async def create_price(
                     status_code=status.HTTP_403_FORBIDDEN,
                     detail="Proof does not belong to current user",
                 )
-    db_price = crud.create_price(db, price=price, user=current_user)
+    db_price = crud.create_price(db, price=price.model_dump(), user=current_user)
```
raphael0202, line 164: Price is supposed to be a `PriceCreate`, no?

tests/test_api.py
raphodn: I honestly don't know, this is more or less a copy-paste of the fastapi tutorials
https://fastapi.tiangolo.com/advanced/testing-database/

pyproject.toml
raphodn: oops, you're right

app/api.py
raphodn: on the crud side we have `Price(**price, user=user)` so I need to check if would work without the `model_dump`

raphodn: If I remove the `model_dump`, I get the following error
```
TypeError: app.models.Price() argument after ** must be a mapping, not PriceCreate
```

raphodn: I moved back the `model_dump` inside the crud file :+1:

tests/test_api.py
raphodn: I'll merge, but these tests will need improvement for sure !

PR #26, opened 2023-11-14
feat: prices GET endpoint

raphodn:
### What

- new `GET` endpoint `/prices`
- no authentication needed
- basic filters

### Todo (futur PRs)

- look at `fastapi-filter` for additional/complex filtering
- and `fastapi-pagination`

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -1,7 +1,9 @@
 import asyncio
 import uuid
+from datetime import date
 from pathlib import Path
 from typing import Annotated
+from typing import List
 
 import requests
 from fastapi import Depends, FastAPI, HTTPException, Request, Response, status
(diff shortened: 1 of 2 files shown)


app/api.py
```diff
@@ -124,6 +126,12 @@ async def authentication(form_data: Annotated[OAuth2PasswordRequestForm, Depends
     raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail="Server error")
 
 
+@app.get("/prices", response_model=List[schemas.PriceBase])
```
raphodn, line 129: @raphael0202 this is a simple v1, but I feel we'll need to have improvements like 
- using [fastapi-filters](https://github.com/arthurio/fastapi-filter) to have more complex filtering
- pagination & limit parameters
- should the endpoint answer a list, or a dict with keys like "data" ?

raphael0202, line 129: we can use directly list[schemas.PriceBase] instead of List, as we're running on Python 3.10

raphael0202: pagination would be good to have indeed!
Better have a dict as a response, so we can add other fields if needed.
like `{"prices": [...]}`

raphodn: in Django Rest Framework the following format is returned
image

raphodn: I found [fastapi-pagination](https://github.com/uriyyo/fastapi-pagination), I'll add it in a follow-up PR

PR #28, opened 2023-11-15
feat: add pagination on GET /prices

raphodn:
### What

- install [fastapi-pagination](https://github.com/uriyyo/fastapi-pagination)
- setup pagination on GET /prices endpoint

### Screenshot

image


app/crud.py
```diff
@@ -66,6 +67,18 @@ def get_prices(db: Session, filters={}):
     return query.all()
 
 
+def get_prices_without_db(filters={}):
```
raphael0202, line 70: It's preferable to avoid having mutable default values for functions in Python. Can you replace it by `filters: dict | None = None` and initialize filters to `{}` in the function if it's None?

raphael0202: And we can rename `get_prices_without_db` to `get_prices_query`, which I find easier to understand (it took me a bit of time to understand what was the point of this)

raphodn: ok !

Issue #31, opened 2023-11-16
Code improvements (tooling, async...)


raphodn:
### Story

For the v1 we went a bit fast, here is a list of known improvements : 
- [ ] schema: add field descriptions (pydantic) & examples
- [ ] background tasks: avoid passing db (see [this discussion](https://github.com/tiangolo/fastapi/discussions/8502))
- [x] tests: teardown data between each test, add more tests
- [x] tests: use PostgreSQL instead of SQLite
- [ ] documentation: improve it (ex: docker), add docstring
- [ ] ci: upgrade release-please to v4

kundan-bhardwaj:
@raphodn I am a newbie in opensource and want to fix this issue can you please elaborate this in detail.

raphodn:
Hi, sorry this issue is outdated (very linked to FastAPI tooling, but we migrated to Django last year)

PR #35, opened 2023-11-16
feat: simple Location model

raphodn:
### What

- new `Location` model
- it's goal is to store metadata fetched from OpenStreetMap, and link it to the Prices (next PR)


app/models.py
```diff
@@ -31,19 +31,44 @@ class User(Base):
     __tablename__ = "users"
 
 
+class Location(Base):
+    id = Column(Integer, primary_key=True, index=True)
+
+    osm_id = Column(BigInteger)
```
raphael0202, line 37: why don't we use osm_id as a primary key directly, as it's an INT?

app/models.py
```diff
@@ -31,19 +31,44 @@ class User(Base):
...
+    osm_type = Column(ChoiceType(PriceLocationOSMType))
```
raphael0202, line 38: Not directly related to this PR, but we could rename PriceLocationOSMType into LocationOSMType (it doesn't add anything to add "Price")

app/schemas.py
```diff
@@ -16,6 +16,26 @@ class UserBase(BaseModel):
...
+class LocationCreate(BaseModel):
+    model_config = ConfigDict(from_attributes=True, arbitrary_types_allowed=True)
+
+    osm_id: int = Field(gt=0)
+    osm_type: PriceLocationOSMType
+
+
+class LocationBase(LocationCreate):
+    id: int
+    osm_name: str | None
```
raphael0202, line 28: It's not mandatory, but it's not clear what each field is about. It's possible to describe the field using pydantic `description` parameter in `Field`.

raphael0202 (approved):

app/models.py
raphodn: I'd rather not rely 100% on OSM. There's probably future locations that will not/never be in OSM. Like online stores, or temporary markets, or even closed locations 🤔...

app/schemas.py
raphodn: i added this to #31

app/models.py
raphodn: done here https://github.com/openfoodfacts/open-prices/pull/35/commits/2637cf778fed400dcf390ac4330b782deefdd7bb

PR #36, opened 2023-11-16
feat: on Price create, create (or get) Location, and link them

raphodn:
### What

- new relationship `Price.location`
- new background task `create_price_location`
- on Price create, we first create (or get if existing) a Location, and then link them


app/tasks.py
```diff
@@ -0,0 +1,15 @@
+from sqlalchemy.orm import Session
+
+from app import crud
+from app.schemas import LocationCreate, PriceBase
+
+
+def create_price_location(db: Session, price: PriceBase):
```
raphael0202, line 7: are `Session` objects thread-safe? As I assume we're using threads to launch a task in the background.

raphael0202 (approved):

app/tasks.py
raphodn: No idea 😅
I used the background tasks example from the fastAPI tutorials..

raphodn: though they do recommend to create a new db session here: https://github.com/tiangolo/fastapi/discussions/8502#discussioncomment-5606921

raphodn: added it to #31

PR #44, opened 2023-11-21
feat: simple Product model

raphodn:
### What

- new `Product` model & schema
- it's goal is to store metadata fetched from OpenFoodFacts, and link it to the Prices (next PRs)


app/enums.py
```diff
@@ -1,6 +1,13 @@
 from enum import Enum
 
 
+class ProductOFFSource(Enum):
```
raphael0202, line 4: There is the `Flavor` enum in the Python SDK just for that

app/schemas.py
```diff
@@ -16,6 +23,22 @@ class UserBase(BaseModel):
...
 
+class ProductCreate(BaseModel):
+    model_config = ConfigDict(from_attributes=True, arbitrary_types_allowed=True)
+
+    code: str
+
+
+class ProductBase(BaseModel):
+    id: int
+    off_source: ProductOFFSource | None
```
raphael0202, line 34: why do you prefix with off? I don't think that the fact that off is the source is relevant, we can probably skip it :)

app/schemas.py
```diff
@@ -16,6 +23,22 @@ class UserBase(BaseModel):
...
+    off_name: str | None
```
raphael0202, line 35: `product_name` maybe? To follow off field names

raphael0202 (approved):

app/enums.py
raphodn: ok awesome !

app/schemas.py
raphodn: if we were to have other sources to get product info ?

app/schemas.py
raphodn: yeah you're right, i'll do the same for `off_product_quantity`

app/schemas.py
raphodn: Ok I'll remove the `off_` prefix :ok_hand:

PR #46, opened 2023-11-22
feat: GET /products/id endpoint to get product details

raphodn:
### What

- new endpoint `GET /products/id` to get product details
- add tests

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -95,7 +95,7 @@ def main_page(request: Request):
 
 
 @app.post("/auth")
-async def authentication(
+def authentication(
     form_data: Annotated[OAuth2PasswordRequestForm, Depends()],
     db: Session = Depends(get_db),
 ):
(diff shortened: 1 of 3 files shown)


app/api.py
```diff
@@ -208,6 +208,17 @@ def get_user_proofs(
     return crud.get_user_proofs(db, user=current_user)
 
 
+@app.get("/products/{product_id}", response_model=schemas.ProductBase)
+async def get_product(product_id: int, db: Session = Depends(get_db)):
```
raphael0202, line 212: We should (almost) always declare our function as normal function (def myfunc), and not coroutine (async def), as we don't support async currently:
see https://fastapi.tiangolo.com/async/

raphael0202 (approved):

app/api.py
raphodn: i removed async everywhere :ok_hand:

PR #49, opened 2023-11-22
feat: fetch product data from OpenFoodFacts

raphodn:
### What

following #45, we want to enrich the newly created Product with data fetched from OpenFoodFacts.

We use [openfoodfacts-python](https://github.com/openfoodfacts/openfoodfacts-python)

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -50,13 +50,13 @@ def fetch_product_openfoodfacts_details(product: ProductBase):
         if response["status"]:
             product_openfoodfacts_details["source"] = Flavor.off
             for off_field in ["product_name", "product_quantity", "image_url"]:
-                if off_field in response[0]:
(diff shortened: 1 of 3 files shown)


app/utils.py
```diff
@@ -26,6 +29,39 @@ def init_sentry(sentry_dsn: str | None, integrations: list[Integration] | None =
...
+        country=Country.world,
+        flavor=Flavor.off,
+        version=APIVersion.v2,
+        environment=Environment.org,
+    )
+    return client.product.get(code)
+
+
+def fetch_product_openfoodfacts_details(product: ProductBase):
+    product_openfoodfacts_details = dict()
```
raphael0202, line 47: just a cosmetic comment, `dict` -> `{}`, it's more concise

raphael0202 (approved):

PR #53, opened 2023-11-23
feat: allow to upload price of barcode-less products

### LINKED ISSUE #47: Add price for a product without EAN code
### Story

Currently only prices with a EAN code are allowed

raphael0202:
Add support for barcode-less product by adding a new `category_tag` field in the `prices` table.

Also:
- fix a recursion error in PriceCreate model, due to date namespace clash
- document all fields of PriceCreate model

Fixes #47.


app/schemas.py
```diff
@@ -84,12 +129,24 @@ def serialize_currency(self, currency: Currency, _info):
             return currency.code
         return currency
 
+    @model_validator(mode="after")
+    def product_code_and_category_tag_are_exclusive(self):
```
raphodn, line 133: :+1:

app/schemas.py
```diff
@@ -56,20 +57,64 @@ class LocationBase(LocationCreate):
...
-    date: date
-    proof_id: int | None = None
+    product_code: str | None = Field(
+        default=None,
+        min_length=1,
+        pattern="^[0-9]+$",
+        description="barcode (EAN) of the product, as a string.",
+        examples=["16584958", "1234567890123"],
+    )
+    category_tag: str | None = Field(
```
raphodn, line 73: in the long terme could this be an enum from the openfoodfacts-python library ?

app/schemas.py
```diff
@@ -56,20 +57,64 @@ class LocationBase(LocationCreate):
...
+        default=None,
+        min_length=3,
+        pattern=r"^[a-z]{2,}:[a-zA-Z\-]+$",
```
raphodn, line 76: we should add tests (not necessarily in this PR if you want to merge quickly)

raphodn (approved):

app/schemas.py
raphael0202: Not really, as the category taxonomy is frequently updated.

app/schemas.py
raphael0202: I tested it locally and it works fine, but we definitely need some :)

raphodn: added here #60

PR #57, opened 2023-11-23
feat: Add image serving

raphael0202:
We need to allow users to fetch images they uploaded.

Two other fixes are included in the PR.

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -1,4 +1,4 @@
-import asyncio
+import time
 import uuid
 from pathlib import Path
 from typing import Annotated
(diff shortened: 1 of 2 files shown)


app/schemas.py
```diff
@@ -33,12 +33,24 @@ class ProductCreate(BaseModel):
 
 class ProductBase(BaseModel):
```
raphodn, line 34: I made a miss here, we could inherit from ProductCreate like the other schemas

raphodn: maybe do this in a seperate PR (inherit + pydantic description/examples)

raphodn: done here : https://github.com/openfoodfacts/open-prices/pull/58
merge 58 first then rebase ? :innocent:

raphael0202: I'm not in favor of multiplying PRs at this stage of the project, as it's more time-consuming. We can do rebase when merging the PR to have clean git histories.

raphael0202: Ok  I didn't understand you wanted to create it yourself ^^". I updated the PR to only keep serving images

raphodn: PRs might be slightly time consuming, but they help when multiple people are working on the project at the same time, avoiding conflitcts, and clarifying what changes were made :)

PR #61, opened 2023-11-23
feat: add labels_tags field

raphael0202:
This field is used to store labels (organic, fairtrade) of products without barcode.


app/schemas.py
```diff
@@ -105,9 +105,25 @@ class PriceCreate(BaseModel):
         This ID must be a canonical category ID in the Open Food Facts taxonomy.
         If the ID is not valid, the price will be rejected.""",
     )
+    labels_tags: list[str] | None = Field(
```
raphodn, line 108: possible to add a regex pattern for `list[str]` fields ? :sweat_smile:

raphodn (approved):
cool ! :label:

app/schemas.py
raphael0202: Not really!
It could be added in the validator, but it's not really necessary, as we check that the label tag is in the taxonomy

PR #62, opened 2023-11-23
feat: add CORS configuration

raphodn:
### What

https://fastapi.tiangolo.com/tutorial/cors/
- add cors configuration in `app.py`
- update `.env`
- update docker config

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -1,6 +1,7 @@
 from enum import Enum
 from pathlib import Path
 
+from pydantic import Field
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 ROOT_DIR = Path(__file__).parent.parent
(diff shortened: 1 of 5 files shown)


app/config.py
```diff
@@ -36,6 +36,7 @@ class Settings(BaseSettings):
     postgres_password: str
     postgres_host: str
     postgres_port: int = 5432
+    cors_allow_origins: list = []
```
raphael0202, line 39: shouldn't you use a Field with a default factory here?
https://docs.pydantic.dev/latest/concepts/fields/#default-values
And you can use list[str]

raphodn: yes you're right, I fixed it here : https://github.com/openfoodfacts/open-prices/pull/62/commits/0fa085bd2cf76d3fe3af70420d03e0ba06766e06

raphodn: not yet a pro of these typings in python ^^

PR #68, opened 2023-11-25
feat: add Proof.type field

raphodn:
### What

New field `Proof.type` allowing to differentiate proofs.
New enum to validate this field, defining a list of possible values : `PRICE_TAG`, `RECEIPT`, `GDPR_REQUEST`, `RETAILER`


app/enums.py
```diff
@@ -5,3 +5,10 @@ class LocationOSMType(Enum):
     NODE = "NODE"
     WAY = "WAY"
     RELATION = "RELATION"
+
+
+class ProofTypeEnum(Enum):
+    PRICE_TAG = "PRICE_TAG"
+    RECEIPT = "RECEIPT"
+    GDPR_REQUEST = "GDPR_REQUEST"
+    RETAILER = "RETAILER"
```
raphael0202, line 14: What kind of proof retailers would submit? Isn't it enough that the data comes from a "verified" account (i.e account with an email address from the retailer)?

raphael0202 (approved):

app/enums.py
raphodn: fyi I was including the COOPs in the RETAILER type
1. it could be just a screenshot with a date ?
2. if we decide to add a price filter on "has proof" or not, it would be better to ask retailers to send a proof, no ?

raphael0202: On Open Food Facts pro platform, we assume the data coming from trusted source to be accurate, as they're the source of the data it makes sense not to ask for an additional proof in my opinion.

> if we decide to add a price filter on "has proof" or not, it would be better to ask retailers to send a proof, no ?

I was considering to add a flag "is_retailer" to the user table, so that we know which prices come from a verified source.

raphodn: ok i removed it !

PR #69, opened 2023-11-27
fix: fix labels_tags field validator

raphael0202:
providing `labels_tags` didn't work, because the validator was broken.


raphodn (approved):
gimme 2 minutes and I'll push a test :pray: 

edit: nevermind i'll do it in another PR

PR #83, opened 2023-12-13
feat: add Price.product_name field

raphodn:
### What

New field `Price.product_name` to store the product name displayed in the stores. We have this info on the proof images, but also on the receipts and in the GDPR requests.

 ### Screenshot

image


app/schemas.py
```diff
@@ -81,13 +81,20 @@ class LocationBase(LocationCreate):
...
     model_config = ConfigDict(from_attributes=True, arbitrary_types_allowed=True)
+
     product_code: str | None = Field(
         default=None,
         min_length=1,
         pattern="^[0-9]+$",
         description="barcode (EAN) of the product, as a string.",
         examples=["16584958", "8001505005707"],
     )
+    product_name: str | None = Field(
```
raphodn, line 92: @raphael0202 in terms of naming, I hesitated with `product_label` and `product_tag`. But there are the existing `category_tag` & `labels_tags` so it could be confusing... or maybe `product_store_label` ? or `product_name_in_store` ?

app/schemas.py
```diff
@@ -81,13 +81,20 @@ class LocationBase(LocationCreate):
...
+        defaul=None,
```
raphael0202, line 93: defaul -> default

app/schemas.py
raphael0202: `product_name` sounds good to be!

app/models.py
```diff
@@ -94,6 +94,7 @@ class Price(Base):
     id = Column(Integer, primary_key=True, index=True)
 
     product_code = Column(String, nullable=True, index=True)
+    product_name = Column(String, nullable=True, index=True)
```
raphael0202, line 97: Do we really want to index it? The database can grow quickly, and it's a string field

raphodn: you're right, fixed here : https://github.com/openfoodfacts/open-prices/pull/83/commits/4b6f20956806fe34f2ff4b52aea0849c93ab1510

PR #85, opened 2023-12-14
feat: add frontend app

raphael0202:
The front-end (the content in the dist folder) must be deployed under /static/app

- use an nginx server as a reverse-proxy
- serve static assets (images, robots.txt, index.html) through nginx
- host OpenAPI docs under /api/docs (instead of /docs)


raphodn (approved):
:raised_hands:

PR #108, opened 2023-12-27
feat: add endpoint to retrieve location by osm data

raphodn:
### What

New endpoint `/locations/osm/<osm_type>/<osm_id>`

- similar to the OpenStreetMap url scheme : https://www.openstreetmap.org/node/6482573185
- will be useful for https://github.com/openfoodfacts/open-prices-frontend/issues/59

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -290,11 +290,30 @@ def get_product(product_id: int, db: Session = Depends(get_db)):
 
 
 @app.get(
+    "/api/v1/locations/osm/{location_osm_type}/{location_osm_id}",
+    response_model=schemas.LocationBase,
+    tags=["Locations"],
+)
+def get_location_by_osm(
+    location_osm_type: str, location_osm_id: int, db: Session = Depends(get_db)
+):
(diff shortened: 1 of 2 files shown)


app/api.py
```diff
@@ -289,12 +289,31 @@ def get_product(product_id: int, db: Session = Depends(get_db)):
     return db_product
 
 
+@app.get(
+    "/api/v1/locations/osm/{location_osm_type}/{location_osm_id}",
```
raphael0202, line 293: Do we really need the OSM type in the URL? Can't the ID uniquely identify the node?

raphodn: no sadly we need the type info, there can be nodes or ways or relations with the same ids

https://wiki.openstreetmap.org/wiki/Elements#Ids
> Users should also be aware that nodes, ways and relations share the same object id space. Any tool which is solely referring to an OSM object by its id will encounter undefined behavior (e.g. Document.getElementById()). OSM Objects must always be referred to both with their object ids and their respective object type (node, way, relation).

PR #110, opened 2023-12-29
feat: add price.origins_tags field

raphael0202:
I also updated the unit test to make them independent of one another: some tests relied on the execution of previous tests.


raphodn (approved):
awesome !

PR #116, opened 2024-01-03
docs: add tutorial + improve main doc page

raphael0202:


docs/README.md
```diff
@@ -1,8 +1,10 @@
 # Open Prices
```
raphodn, line 1: Maybe you wanted to edit the root README instead ? (and we could delete this docs/README ?)

docs/assets/img/proof-examples.png
teolemon, line 1: What's wrong with the second example ?

teolemon: Also, do you have a source template you can commit ?

raphodn: Same for the second image 😅
Maybe put a blurred image ? And/or a badly cropped image ?

PR #119, opened 2024-01-05
feat: import product DB every day

raphael0202:
Add:
- CLI with Typer
- CLI command to import/update products in DB
- scheduler (APScheduler)
- scheduled job everyday to import/update new products in DB


app/models.py
```diff
@@ -44,6 +44,7 @@ class Product(Base):
     product_quantity = Column(Integer)
     brands = Column(String)
     image_url = Column(String)
+    unique_scans_n = Column(Integer, nullable=True)
```
raphodn, line 47: we could default this to 0 ?

app/scheduler.py
```diff
@@ -0,0 +1,24 @@
...
+def import_product_db_job():
+    db = session()
+    import_product_db(db=db)
+
+
+def run():
+    scheduler = BlockingScheduler()
+    scheduler.add_executor(ThreadPoolExecutor(20))
+    scheduler.add_jobstore(MemoryJobStore())
+    scheduler.add_job(
```
raphodn, line 21: so if we deploy this it will run every morning at 6AM is that right ?

raphodn: prod only, or also preprod ?

PR #121, opened 2024-01-05
feat: update filter library

raphael0202:
Switch from fastapi-filter to https://github.com/uriyyo/fastapi-filters that allow to put null values last when performing a sort.
This is necessary when sorting by `unique_scans_n`, as we don't want null values first (we want the most popular first).
This is a breaking change in the API, as we cannot exactly mimick the parameters of fastapi-filter.


pyproject.toml
```diff
@@ -42,7 +42,7 @@ sqlalchemy = "~2.0.23"
 sqlalchemy-utils = "~0.41.1"
 uvicorn = "~0.23.2"
 fastapi-pagination = "^0.12.12"
-fastapi-filter = "^1.0.0"
+fastapi-filters = "~0.2.9"
```
raphodn, line 45: Ah sucks that we have to change filtering library.. Is it a known problem on their side with a tracked issue ? The new library has less users and activity it seems.

raphodn: I'll check tomorrow when I get back behind the computer :)

raphodn: I ended up doing a PR #122, to seperate at least the new endpoint and this PR that replaces the filtering library. let me know i can take care of the rebase :pray:

raphodn: an alternative would be to init unique_scans_n to 0

raphael0202:
Closing, as it's not required anymore.

PR #131, opened 2024-01-11
feat: add proof object in price response

raphael0202:


raphodn:
@raphael0202 is there a public url to actually display the proof in the frontend ?

I tried 
```
https://prices.openfoodfacts.org/0002/qU59gK8PQw.webp
https://prices.openfoodfacts.org/app/0002/qU59gK8PQw.webp
https://prices.openfoodfacts.org/api/0002/qU59gK8PQw.webp
https://prices.openfoodfacts.org/public/0002/qU59gK8PQw.webp
```

PR #150, opened 2024-01-15
feat: paginate, filter & order proofs

raphodn:
### What

Improve the current GET `/proofs` endpoint with : 
- pagination
- filter on `type` (reminder: also a default filter on `owner` as the authenticated user)
- sort

Add tests


raphael0202 (approved):
:+1: 
There is a merge conflict to resolve

PR #163, opened 2024-01-25
feat: add price.price_is_discounted field

raphodn:
### What

Following #130, most of the times the user does not know the full price.

This new boolean field allows to : 
- only pass the info that the price is discounted (without the full price)
- easily filter on prices with a discount

Update schema, add validation & tests


app/schemas.py
```diff
@@ -198,6 +198,11 @@ class PriceCreate(BaseModel):
         description="price of the product, without its currency, taxes included.",
         examples=["1.99"],
     )
+    price_is_discounted: bool | None = Field(
```
raphael0202, line 201: `bool` instead of `bool | None` no?

raphodn: I wasn't sure what would happen if the field wasn't provided on create. I'll double check and fix 👍

raphodn: ok removed :ok_hand:

PR #178, opened 2024-01-31
feat: allow to filter based on categories_tags

raphael0202:
poke @raphodn


raphodn:
thanks @raphael0202 ! so you're saying the JSONB is not compatible with sqlite ?
- i'm going to look into postgresql array : https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#array-types
- maybe split the PR into 2 or 3 : new category_tags field ; sync category_tags (maybe there's nothing to do ?) ; add filtering

raphael0202:
> So you were saying the JSONB is not compatible with sqlite and makes the test fail ?

Yes you can run the unit tests to see the error message.

> sync category_tags

I added the `categories_tags` field to OFF_FIELDS so it's synced as well

raphodn:
- sqlite was removed in https://github.com/openfoodfacts/open-prices/pull/201
- the categories_tags will be added as a PostgreSQL Array type in https://github.com/openfoodfacts/open-prices/pull/192
- will need to open a third PR for filtering

PR #179, opened 2024-01-31
feat: new DELETE prices endpoint

raphodn:
### What

New endpoint `/prices/{price_id}` to allow a user to delete one of it's price
The user must be authenticated + the price id must exist + the user must be the owner of the price

Also update `price_count` fields

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -329,7 +329,7 @@ def create_price(
     """
     # check if we have a proof_id provided
     if price.proof_id is not None:
-        db_proof = crud.get_proof_by_id(db, proof_id=price.proof_id)
+        db_proof = crud.get_proof_by_id(db, id=price.proof_id)
         if db_proof is None:
             # No proof exists with this id
             raise HTTPException(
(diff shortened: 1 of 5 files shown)


app/api.py
```diff
@@ -339,19 +339,63 @@ def create_price(
...
+            detail=f"Price with code {price_id} not found",
+        )
+    # Check if the price belongs to the current user
+    if db_price.owner != current_user.user_id:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Price does not belong to current user",
+        )
+    # update counts
+    background_tasks.add_task(tasks.decrement_user_price_count, db, user=current_user)
```
raphael0202, line 387: Why do you do this as a background task? I would rather have it in the endpoint, so that the full operation is atomic.

app/crud.py
```diff
@@ -239,6 +250,17 @@ def increment_product_price_count(db: Session, product: ProductFull):
     return product
 
 
+def decrement_product_price_count(db: Session, product: ProductFull):
+    """Decrement the price count of a product.
+
+    This is used to keep track of the number of prices linked to a product.
+    """
+    product.price_count -= 1
+    db.commit()
```
raphael0202, line 259: Regarding the atomicity of the transaction, we should avoid adding db.commit, and only commit at the end of the transaction (in peewee we can use a context manager to take care of this, but I don't know how it works in SQLAlchemy).

app/crud.py
```diff
@@ -291,6 +317,12 @@ def set_price_location(db: Session, price: PriceFull, location: LocationFull):
     return price
 
 
+def delete_price(db: Session, db_price: PriceFull):
```
raphael0202, line 320: can you declare -> bool as a return type ?

app/api.py
raphodn: ok fixed ! i removed the background tasks, and did 1 single crud method / commit instead :)

Issue #182, opened 2024-02-01
Add support for moderators


raphael0202:
Moderators are special users that will be allowed to delete/update prices or proofs of other users.
This is necessary to ensure database quality.

Tasks:
- [x] create a new `is_moderator` field in the `users` table : #183
- [x] allow moderators to edit proofs : #235 
- [x] allow moderators to delete proofs (without prices) : #197
- [x] allow moderators to edit prices : #259 
- [x] allow moderators to delete prices : #260

teolemon:
FYI https://github.com/openfoodfacts/openfoodfacts-server/pull/9918

PR #183, opened 2024-02-02
feat: new user.is_moderator field

alvarobertin:
### What
- new feature of is_moderator field that lets users modify and add proofs that are not theirs.

### Part of 

#182  issue


app/models.py
```diff
@@ -28,6 +28,7 @@ class User(Base):
     user_id = Column(String, primary_key=True, index=True)
     price_count = Column(Integer, nullable=False, server_default="0", index=True)
     created = Column(DateTime(timezone=True), server_default=func.now())
+    is_moderator = Column(Boolean, nullable=False, default=False)
```
raphodn, line 31: `server_default` is used elsewhere instead of `default`. maybe stick instead to what is currently done ?

app/crud.py
```diff
@@ -159,6 +159,35 @@ def delete_session(db: Session, session_id: int) -> bool:
     return results.rowcount > 0
 
 
+def update_user_moderator(db: Session, user_id: str, is_moderator: bool) -> bool:
```
raphodn, line 162: this is used only for tests, right ?

app/crud.py
```diff
@@ -159,6 +159,35 @@ def delete_session(db: Session, session_id: int) -> bool:
...
+    """
+    db_user = get_user_by_user_id(db, user_id=user_id)
+    if db_user:
+        db_user.is_moderator = is_moderator
+        db.commit()
+        return True
+    return False
+
+
+def is_user_moderator(db: Session, user_id: str) -> bool:
```
raphodn, line 178: this does not seem to be used ?

app/schemas.py
```diff
@@ -35,6 +35,7 @@ class UserBase(BaseModel):
 
     user_id: str
     price_count: int = 0
+    is_moderator: bool = False
```
raphodn, line 38: I'm not sure if we should make this info available in the api :thinking:

raphodn (commented):
Thanks for the PR ! a few comments

app/models.py
alvarobertin: I tried, but it does not work when creating an user with crud.create_user(). I'm going to pass the  is_moderator field as an optional parameter for the crud.create_user() func.

app/crud.py
alvarobertin: Should I remove it? 😄

app/crud.py
alvarobertin: True, removed

app/crud.py
raphodn: no it's fine i guess ^^

raphodn (commented):
looks good, just the alembic migration missing ?

raphodn (approved):

PR #191, opened 2024-02-05
test: use PostgreSQL instead of SQLite

raphodn:
### What

Run the tests against the same type of database as the app (PostgreSQL)


raphodn:
done in #201 :tada:

PR #192, opened 2024-02-05
feat: new Product.categories_tags array field

raphodn:
### What

New `Product.categories_tags` field.
- use PostgreSQL Array field type
- add to schema (will be returned by the API)
- add to OFF sync

### Extra info

- why `server_default="{}"` ? see https://stackoverflow.com/questions/30933266/empty-array-as-postgresql-array-column-default-value


app/utils.py
```diff
@@ -38,6 +38,7 @@ def init_sentry(sentry_dsn: str | None, integrations: list[Integration] | None =
     "product_name",
     "product_quantity",
     "product_quantity_unit",
+    "categories_tags",
```
raphodn, line 41: I added the field in the OFF sync, but i was wondering if we should get the raw data or filter it a bit (for instance only keep the `en:` categories ?

example : https://world.openfoodfacts.org/api/v2/product/8001505005592.json
```
"categories_tags":["en:breakfasts","en:spreads","en:sweet-spreads","fr:pates-a-tartiner","en:hazelnut-spreads","en:chocolate-spreads","en:cocoa-and-hazelnuts-spreads"]
```

raphael0202: Sorry I missed this comment. I think we can add all categories, some categories don't start with `en:` and are however valid categories with respect to the category taxonomy.

raphodn: Ok, but what is the default behavior ? If there is an english translation, is the category stored as en: by default ? Or can there be both en & non en categories ?

PR #194, opened 2024-02-07
feat: new Product.product_quantity_unit field

raphodn:
### What

Thanks to https://github.com/openfoodfacts/openfoodfacts-server/pull/9608 there is now a new field in OFF : `product_quantity_unit` (we already stored `product_quantity`)

- added the field to the `Product` model
- updated the sync with OFF
- added the field to the schema (will be returned by the API)


raphael0202 (approved):
LGTM!

Issue #180, opened 2024-02-01
Add an endpoint to delete a proof


raphael0202:
Currently, we can't delete a proof once uploaded.

Endpoint: `DELETE /api/v1/proofs/:proof_id`

This endpoint should deny deleting proofs if:
- some prices are associated with it
- if the user is not the owner of the proof

PR #197, opened 2024-02-07
feat: new DELETE proofs endpoint

alvarobertin:
### What
- New Endpoint: `DELETE /api/v1/proofs/:proof_id`.
- Only lets you delete if there are no prices associated with it 
-  Only lets you delete if  if you are the owner (if you are moderator you can delete it)

### Part of 
- #180


app/crud.py
```diff
@@ -438,6 +439,19 @@ def increment_proof_price_count(db: Session, proof: ProofFull):
     return proof
 
 
+def delete_proof(db: Session, db_proof: ProofFull) -> bool:
+    # we delete the image of the proof
+    file_path_obj = Path(db_proof.file_path)
+    # Check if the file exists
+    if file_path_obj.exists():
+        # remove the file
+        file_path_obj.unlink()
```
raphodn, line 448: @raphael0202 ok for you to hard delete proof images ?

raphodn (approved):
great !

app/crud.py
raphael0202: yes, no issue on my side :)

PR #201, opened 2024-02-09
ci: run integration tests and unit tests in docker. remove sqlite

raphael0202:
This way, we can test against PostgreSQL database. In the commit, we also:

- removed labels_tags__like and origins_tags__like filters. These filters triggers an error in PostgreSQL, as we're using a JSONB field type, which does not support LIKE queries (this bug was in production too, see https://openfoodfacts.sentry.io/issues/4911496304/?project=4506270533091328&query=LIKE&referrer=issue-stream&sort=freq&statsPeriod=90d&stream_index=0)
- fixed some integration tests to make it work with PostgreSQL

### DIFF
--- a/.github/workflows/quality-check.yml
+++ b/.github/workflows/quality-check.yml
@@ -1,0 +1,42 @@
+name: Open Prices unit and integration tests
+
+on:
+  push:
+    paths:
+      - "**/app/**"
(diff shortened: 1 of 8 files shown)


raphodn (approved):
Yesss finally :pray: :100:

PR #203, opened 2024-02-10
build: standardize and upgrade python version

ArturLange:
### What
This PR upgrades python version to 3.12. There should be no breaking changes.
It also aims to fix all inaccuracies with python versions (e.g. `Dockerfile` vs. `pyproject.toml`)

### Part of 
- #31


INSTALL.md
```diff
@@ -2,7 +2,7 @@
 
 ## Prerequisites
 
-- Python 3.10 (lower version may be OK, but untested)
+- Python 3.12 (lower version may be OK, but untested)
```
raphodn, line 5: Not sure what is the best strategy here
- Python 3.12 was only released in [Octobre 2023](https://www.python.org/downloads/release/python-3120/), I don't believe many developers have it yet on their laptop
- For instance I run Ubuntu LTS 22.04, and it comes with Python 3.10

What's the good practice ? To have an app that works with the lowest/most common version of Python ? Or with the latest, with the risk of causing problems for many developers ?

Interested to have some opinions, I would have standardized with the lowest version for instance

ArturLange: If it was a package that users install on their computers, then we should definitely aim for the widest range of versions and test for them all.
But this is the backend of an web application, end users don't need to have it on their computers.
For the developers - it is dockerized and with virtualenvs and tools like `pyenv` and `poetry` it shouldn't be much of a problem. 
If we depend on the default system version of Python though (which I'm pretty sure we shouldn't do), we're gonna have a bad time - right now it may be Python 3.10 from 2.5 years ago but it will get worse.

raphael0202: From experience, it's not the best idea to use the latest version available as there may be some bugs in the libraries/framework we use with the new version. I would rather stay with 3.11, especially as there are no feature in 3.12 we need right now.

ArturLange: I checked and was surprised to see you're right - most of the features I want are in 3.11.
I will change it to 3.11 👍

raphodn: which feature ? :upside_down_face:

raphael0202:
@ArturLange pre-commit hook fails for your latest commit, can you check this?

raphael0202 (approved):
LGTM, thank you!

PR #210, opened 2024-02-16
perf: enable gzip encoding

ArturLange:
### What
Enable gzip encoding to improve web app performance.

Inspired by [these findings](https://github.com/openfoodfacts/open-prices-frontend/issues/88#issuecomment-1941359326).


raphael0202:
Thank you!

PR #212, opened 2024-02-20
perf: enable cache for static files in nginx

ArturLange:
### What
Enable cache for static files in nginx.
Add max possible age for them as files built by `vite` have a hash and will be invalidated after each change


raphael0202 (approved):
LGTM, thank you!

PR #226, opened 2024-02-23
refactor: Simplify api.py (move endpoints to their own routers)

ArturLange:
### What

Move endpoints to different directory to clean up `api.py` 

See https://fastapi.tiangolo.com/tutorial/bigger-applications/


app/views/auth.py
raphodn, line 1: Thanks @ArturLange for starting this refactoring work. I have a question on the approach : why the `views` folder naming ? did you use a specific tutorial as reference ?

raphodn: I did a bit of research and I found 

1)  the fasapi doc : https://fastapi.tiangolo.com/tutorial/bigger-applications/
which favors naming them `routes` + extracting the route prefix

2) a reddit question : that favors the Django approach of having 1 folder per app

raphodn: being a Django dev (and slightly regretting not having chosen Django for this backend ^^), I kinda favor that approach, though we're probably not there yet in terms of complexity. so maybe go with the "official" routing approach for now ?

ArturLange: I agree with the `routes` point, I can change that. For the "one folder per app" thing - AFAIK in Django "app" isn't what we usually mean - one web application can have many "apps" and I don't think that applies in FastAPI.
I'm against having `api.py` file with thousands of lines of code - it's hard to look at and it makes it harder to change something without creating conflicts with other people.

raphodn: Thanks for the answer :)

Actually the Django approach would indeed mean multiple folders *inside* the app folder.

Same as you, the api.py has got too long and it's a good time to refactor 👍

raphodn: @ArturLange done ? there's a CI that is failing :)

app/routes/auth.py
```diff
@@ -0,0 +1,133 @@
...
+        if session:
+            return crud.update_session_last_used_field(db, session=session)
+    raise HTTPException(
+        status_code=status.HTTP_401_UNAUTHORIZED,
+        detail="Invalid authentication credentials",
+        headers={"WWW-Authenticate": "Bearer"},
+    )
+
+
+@auth_router.post("", tags=["Auth"])
```
raphodn, line 49:
```suggestion
@auth_router.post("")
```
no need to repeat the `tags` info now that it is defined at the top :)

app/views/auth.py
raphodn: Ok I did some additional cleanup :+1:

ArturLange: Thank you!
Sorry for the delay on my side

PR #228, opened 2024-02-23
feat: new GET proofs by id endpoint

raphodn:
### What

New endpoint `/proofs/{proof_id}` to allow an owner (or moderator) to fetch its proof

### Why ?

So that we can have in the frontend a proof page (+ proof prices + button to add a new price)


raphodn:
Merging to test in preprod.

To be discussed if we let any user get any proof data ? I filtered on owner or moderator (like the current proof list endpoint)

PR #235, opened 2024-03-02
feat: add PATCH endpoint to update proof

### LINKED ISSUE #236: Add an endpoint to update a proof
### Story

Similar to #181, but for proofs

Allow owners to change the following fields : 
- `type`
- `is_public`

dq18:
### What

- Added a PATCH /api/v1/proofs/{id} endpoint to allow update of boolean `is_public` and to change `type`

### Why

- `is_public` can only be set to `False` at creation/upload. Now user can correct this afterwards.
- Added `type` as patchable. In the frontend, a proof  `type` is automatically set depending on the "Add Price" mode selected. In the future, proof creation and price addition will be decoupled and a page dedicated to proof(s) upload in frontend should come soon. User might not set the right `type`.

Closes #236

### DIFF
--- a/app/crud.py
+++ b/app/crud.py
@@ -24,9 +24,9 @@ from app.schemas import (
     ProductCreate,
     ProductFilter,
     ProductFull,
+    ProofBasicUpdatableFields,
     ProofFilter,
     ProofFull,
-    ProofBasicUpdatableFields,
     UserCreate,
 )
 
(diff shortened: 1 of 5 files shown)


app/crud.py
```diff
@@ -26,6 +26,7 @@
     ProductFull,
     ProofFilter,
     ProofFull,
+    ProofBasicUpdatableFields,
```
raphodn, line 29: there's some CI errors (pre-commit), I'll fix them before merging :pray:

raphodn (approved):
perfect :heart_eyes:

app/crud.py
raphodn: https://github.com/openfoodfacts/open-prices/blob/main/CONTRIBUTING.md#install-pre-commit-hooks

raphodn: Ok fixed here : https://github.com/openfoodfacts/open-prices/pull/235/commits/725f80b7022f4ca6d411be7fc502a36015199646

I ran `poetry run pre-commit run --all-files`

PR #240, opened 2024-03-04
fix: use LocationFull for fixtures

vjousse:
### What

Locations tests are commented because they fail.

### Fixes bug(s)

`LocationCreate` used to create fixtures is not a complete object so some fields are empty in the DB when using the paginate query. By using `LocationFull` we create a full entry in the DB (with an `osm_name`) that can be queried.

### DIFF
--- a/app/crud.py
+++ b/app/crud.py
@@ -502,9 +502,7 @@ def get_location_by_osm_id_and_type(
     )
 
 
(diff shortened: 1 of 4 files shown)


raphodn (approved):
Thanks a lot for your help :pray:

PR #242, opened 2024-03-06
fix: setting cookie on the response

ArturLange:
### What
While working on typing I found [this fragment of code what seems to be causing errors and if used correctly do nothing](https://github.com/openfoodfacts/open-prices/blob/main/app/routers/auth.py#L100):
```python
if r.status_code == 200:
    # form_data.username can be the user_id or the email, so we need to
    # fetch the user_id from the response
    # We also need to lowercase the user_id as it's case-insensitive
    user_id = r.json()["user_id"].lower().strip()
    token = create_token(user_id)
    session, *_ = crud.create_session(db, user_id=user_id, token=token)
    session = crud.update_session_last_used_field(db, session=session)
    # set the cookie if requested
    if set_cookie:
        # Don't add httponly=True or secure=True as it's still in
        # development phase, but it should be added once the front-end
        # is ready
        response.set_cookie(key="opsession", value=token) # <================= this one
    return {"access_token": token, "token_type": "bearer"}
```

What I think that was meant to do is add the "Set-Cookie" header on the response that is returned in that view so I changed the code to do just that.


raphael0202:
Thank you for your PR!

Issue #243, opened 2024-03-06
Product synchronization issue


raphael0202:
This product (https://world.openfoodfacts.org/product/7792052075169/cereal-de-arroz-con-chocolate-rock-choc) was uploaded on Open Food Facts on 2024-02-26, but it's not synced in Open Prices database (as of 2024-03-06).

raphodn:
the product in Open Prices : https://prices.openfoodfacts.org/app/products/7792052075169

and the JSON returned
```
{
    "code": "7792052075169",
    "id": 3081469,
    "source": null,
    "product_name": null,
    "product_quantity": null,
    "product_quantity_unit": null,
    "categories_tags": [],
    "brands": null,
    "brands_tags": [],
    "labels_tags": [],
    "image_url": null,
    "nutriscore_grade": null,
    "unique_scans_n": 0,
    "price_count": 1,
    "created": "2024-02-26T02:43:16.810574Z",
    "updated": null
}
```

dq18:
Seems to me that:
https://github.com/openfoodfacts/open-prices/blob/21c27930250793092c9859de24430342626db59e/app/tasks.py#L140
is the blocking point. Newly created products in Open Prices not yet in OFF will have their source put to `null` by default

raphodn:
It seems that the PR #250 did not fix the problem.

For example, today a price for `3384630300103` was added ([here](https://prices.openfoodfacts.org/app/products/3384630300103)), the product did not exist in OFF yet (added afterwards [here](https://fr.openfoodfacts.org/produit/3384630300103/petite-tomme-de-chevre-giralp)), and the API returns `source=off`

raphodn:
fixed in #257

PR #250, opened 2024-03-13
fix(products): update products added first in Open Prices

dq18:
### What
Products first entered in Open Prices were not updated once added in OFF. Their "source" is set to None/null.

### Potential issue
- if user enters a product with barcode that is not a food product but has the same barcode as a food product (is that possible?)

### Fixes bug(s)
- #243

### DIFF
--- a/app/tasks.py
+++ b/app/tasks.py
@@ -133,10 +133,13 @@ def import_product_db(db: Session, batch_size: int = 1000) -> None:
             item["image_url"] = generate_openfoodfacts_main_image_url(
                 product_code, images, product["lang"]
             )
+            item["source"] = Flavor.off
             item = normalize_product_fields(item)
             execute_result = db.execute(
                 update(Product)
                 .where(Product.code == product_code)
+                # Update the product if it is part of OFF
+                # or if it has no source (created in Open Prices before OFF)
                 .where(
                     or_(
                         Product.source == Flavor.off,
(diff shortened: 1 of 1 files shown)


app/tasks.py
```diff
@@ -146,7 +151,12 @@ def import_product_db(db: Session, batch_size: int = 1000) -> None:
                         Product.updated == None,  # noqa: E711, E501
                     )
                 )
-                .values(**item)
+                .values(
+                    source=Flavor.off
```
raphodn, line 155: `source=item.get("source", Flavor.off)`

app/tasks.py
```diff
@@ -137,7 +137,12 @@ def import_product_db(db: Session, batch_size: int = 1000) -> None:
             execute_result = db.execute(
                 update(Product)
                 .where(Product.code == product_code)
-                .where(Product.source == Flavor.off)
+                .where(
+                    or_(
+                        Product.source == Flavor.off,
```
raphodn, line 142: maybe add a comment to explain why we have these 2 usecases ?

app/tasks.py
raphodn: actually the `item.get("source")` will always be None, as "source" is not in OFF_FIELDS

raphodn: so i would add this above in the "item" creation section
`item["source"] = Flavor.off`

dq18: Ideed. I wanted a conditional in the values on `Product.source is None` but that did not work out so I thought `item` would do. Thought about using a case statement but not sure how that works in SQLAlchemy in the update method.

app/tasks.py
dq18: Added

PR #272, opened 2024-04-17
feat: filter prices by product fields (source, category) (WIP)

raphodn:
### What

WIP


raphodn:

PR #288, opened 2024-05-06
refactor(sync): log extra info

raphodn:
### What

Add some logging during the sync task, to better understand how many products are available in the dumps

### DIFF
--- a/app/tasks.py
+++ b/app/tasks.py
@@ -71,13 +71,21 @@ def import_product_db(
     """
     logger.info(f"Launching import_product_db ({flavor})")
     existing_codes = set(db.execute(select(Product.code)).scalars())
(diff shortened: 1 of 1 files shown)


app/tasks.py
```diff
@@ -71,13 +71,21 @@ def import_product_db(
...
+        f"OP: number of existing codes from {flavor}: {len(existing_codes_flavor)}"
+    )
+
     dataset = ProductDataset(
         flavor=flavor,
         dataset_type=DatasetType.jsonl,
         force_download=True,
         download_newer=True,
     )
+    logger.info(f"Dataset size: {dataset.count()}")
```
raphodn, line 88: the `.count()` command seems to take a looong time for the OFF dataset :thinking:

raphodn:

PR #299, opened 2024-05-21
fix(data): fix daily data import script

raphodn:
### What

In #119 we added the `import_product_db` function that added or updated products from the OFF database (and in #264 we also sync products from OBF, OPF & OPFF). 

But some recently added products in OBF or OPF did not seem to appear the next day in OP.

The issue seemed to be the frequency of `db.commit()` in the script, done in batch but it would forget the very last product changes.


raphael0202 (approved):
Great catch, thank you!

PR #300, opened 2024-05-22
feat: fixtures to fill data in a local environment

Shadows97:
### What
- Add fixtures data and import script for local environment


CHANGELOG.md
raphodn, line 1: Did you find a way to avoid having these file changes ? it would reduce drastically the number of "Files changed", as well as the change history of the related files

Shadows97: i think that i will create new branch and edite the good file who need update

PR #303, opened 2024-05-27
feat: fixtures to fill data in a local environment

Shadows97:
### What
- Add fixtures to fill data in a local environment

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -46,6 +46,12 @@ The run the migration of the database with
 make migrate-db
 ```
 
+The load the fixtures with
+
+```sh
+make load-fixtures
+```
+
 Congrats, you can now contribute to the codebase :tada:
 
 ## Run locally
--- a/fixtures/data.yaml
+++ b/fixtures/data.yaml
@@ -1,0 +1,1073 @@
+users:
+  - user_id: "john_doe"
+    price_count: 15
+    is_moderator: true
+  - user_id: "jane_smith"
+    price_count: 8
+    is_moderator: false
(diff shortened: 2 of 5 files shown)


fixtures/import_data.py
```diff
@@ -0,0 +1,72 @@
+import yaml
+from openfoodfacts import Flavor
+from sqlalchemy.orm import Session
+
+from app.db import Base, engine
```
raphodn, line 5: when i tried to run it locally - `poetry run python fixtures/import_data.py` - i get the following error
```
Traceback (most recent call last):
  File "....../open-prices/fixtures/import_data.py", line 5, in <module>
    from app.db import Base, engine
ModuleNotFoundError: No module named 'app'
```

fixtures/data.yaml
```diff
@@ -0,0 +1,1073 @@
+users:
+  - user_id: "john_doe"
```
raphodn, line 2: just to be sure, these users are fake so it won't be possible to login with them, right ? or do you have a way to bypass the OFF login ?

INSTALL.md
```diff
@@ -46,6 +46,12 @@ The run the migration of the database with
 make migrate-db
 ```
 
+The load the fixtures with
```
raphodn, line 49:
```suggestion
You can load some fixtures with
```

raphodn:
Closing as this does not work anymore, following the migration from FastAPI to Django (#366)..
Feel free to open a new PR that is Django-compatible :innocent:

PR #324, opened 2024-06-16
refactor(proofs): new Proof.updated field

raphodn:
### What

When updating a proof, store the timestamp in the `updated` field


raphodn:
@raphael0202 there seems to be a docker error when running tests ?

PR #349, opened 2024-06-29
feat(proofs): improve schema and add validation for create & update

raphodn:
### What

Following #343, and similar to #342 (for prices)

Add validations on proof create & update : 
- the recently added fields (`location_osm_id`, `location_osm_type`, `date`, `currency`) are now mandatory if the proof type is `PRICE_TAG` or `RECEIPT`
- for both proof create & proof update endpoints
- update tests

### DIFF
--- a/app/crud.py
+++ b/app/crud.py
@@ -24,9 +24,9 @@ from app.schemas import (
     ProductCreate,
     ProductFilter,
     ProductFull,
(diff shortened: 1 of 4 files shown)


raphael0202 (approved):
Great! Looks good to me

raphodn:

PR #361, opened 2024-07-16
chore: add home_cache volume

raphael0202:
to prevent docker overlay due to ~/.cache


raphodn:
+ fix : https://github.com/openfoodfacts/open-prices/commit/c6e566ad99d5229aa0a532e7535a65803365f877

Issue #355, opened 2024-07-14
Rewrite the app in Python Django :rocket: 


raphodn:
I've been thinking lately... 

### Why ?

Having a couple of years of experience in Python Django, but only a few months now with Python FastAPI, I'm struggling to develop the backend as fast and solid as I wish..

### Advantages of moving from FastAPI to Django

- a shell !
- ORM much better than sqlalchemy (joins, prefetch)
- API : use DRF : better serialization, better filtering (instead of https://github.com/arthurio/fastapi-filter)
- an admin interface :heart_eyes: 
- faster development, more solid backend, better manage the app's growth (my feeling is that FastAPI is good for prototyping/small projects, but harder to manage for bigger projects, with many dependencies (alembic, sqlalchemy, filter...) that are far from on-par with Django)
- 

### Disadvantages

- time to rewrite the app
- no typing
- API slower ? how slower ?
- cron managed by the server ? or django-q !
- 

### Extra info

- the initial POC - https://github.com/raphodn/open-prices-poc - was written in Django

raphodn:
Todo : 
- [ ] django setup
- [ ] models
- [ ] api & serializers
- [ ] authentication
- [ ] cron, tasks
- [ ] devops, docker

raphodn:
Which cron ?

|Name|Pros|Cons|Stars|
|---|---|---|---|
|[django-crontab](https://github.com/kraiz/django-crontab)|- simple|- only manages crons<br>- no admin integration|~800|
|[django-celery](https://github.com/celery/django-celery/)|- Celery is famous|- cannot use DB as broker. need to install Redis dependency|~1500|
|[django-q](https://github.com/Koed00/django-q)|- manage tasks, cron...<br>- can use DB as broker<br>- admin integration||~1800|

raphodn:
@monsieurtanuki deployed in prod !

raphodn:
created a discussion here with a quick todo of next/upcoming issues : https://github.com/openfoodfacts/open-prices/discussions/421

PR #376, opened 2024-08-01
feat(Django): daily sync of products from OFF

raphodn:
### What

Installed `django-q2`
```
poetry add django-q2 blessed croniter
```
- with 2 extra packages
- using the DB as message broker
- why this package ? See #355 for a comparison

### Commands

Create scheduled task (how to create automatically ?)
```
from django_q.models import Schedule

Schedule.objects.create(func="open_prices.common.openfoodfacts.import_product_db_full", schedule_type=Schedule.CRON, cron="0 15 * * *")
```

Sync product flavor manually
```
// python manage.py shell_plus
import openfoodfacts
from open_prices.common import openfoodfacts as common_openfoodfacts

common_openfoodfacts.import_product_db(flavor=openfoodfacts.Flavor.opf)
```

### DIFF
--- a/open_prices/common/openfoodfacts.py
+++ b/open_prices/common/openfoodfacts.py
@@ -1,6 +1,19 @@
+import datetime
+
 import requests
+import tqdm
 from django.conf import settings
-from openfoodfacts import API, APIVersion, Country, Environment, Flavor
+from django.db.models import Q
+from django.utils import timezone
+from openfoodfacts import (
+    API,
+    APIVersion,
+    Country,
+    DatasetType,
+    Environment,
+    Flavor,
+    ProductDataset,
+)
 from openfoodfacts.images import generate_image_url
 from openfoodfacts.types import JSONType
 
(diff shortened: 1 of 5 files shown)


open_prices/common/openfoodfacts.py
```diff
@@ -107,3 +120,133 @@ def get_product_dict(product, flavor=Flavor.off) -> JSONType | None:
...
+                    Q(source_last_synced__lt=product_source_last_modified)
+                    | Q(source_last_synced=None)
+                )
+            )
+            if existing_product_qs.exists():
+                products_to_create_or_update.append(Product(**product_dict))
+                updated_count += 1
+
+        # update the database regularly
+        if (
```
raphodn, line 232: in the previous script we had `if buffer_len % batch_size == 0:`
but `0 % 1000 == 0` so it seems like it was commiting the db transactions for EVERY product and not in batch :sweat_smile:

PR #433, opened 2024-09-10
refactor: Allow to load dotenv and exclude it from git track

BhuwanPandey:
- Added .env.example files that contain content as .env
- Exclude .env file from git track
- Added python-dotenv package to load enviroment variables from .env


raphodn:
hI @BhuwanPandey !
Even though our documentation mentions running the project locally, the preferred method remains docker. We don't want to support to much the alternative methods. What is the motivation behind this PR ? Is it to deploy the app on a server without Docker ? If only to run locally, you could look at `direnv` (a bit of docs by OFF [here](https://openfoodfacts.github.io/openfoodfacts-server/dev/how-to-use-direnv/))

BhuwanPandey:
> hI @BhuwanPandey !
> 
> Even though our documentation mentions running the project locally, the preferred method remains docker. We don't want to support to much the alternative methods. What is the motivation behind this PR ? Is it to deploy the app on a server without Docker ? If only to run locally, you could look at `direnv` (a bit of docs by OFF [here](https://openfoodfacts.github.io/openfoodfacts-server/dev/how-to-use-direnv/))
> 
> Don't hesitate to open an issue and we can move this discussion there.

I run this project with and without docker.
Without docker, When i tried, I found that env vars wasnot loading on project. sothat a reason, I added python-dotenv to load env on project cause I think that most django project follow same pattern to load env. If i need to use direnv , i think it will not be project specific. I have to download it globally on my system.

BhuwanPandey:
If we don't want to add extra alternative like this, then you can close this issue. but I think it will be better to rename .env to .env.examples and untrack .env file Isn't it?

while running docker compose up , two of image (api-1 and schedular-1 ) are not running it say
(docker-entrypoint.sh: not found)
```
postgres-1   | sh: locale: not found
postgres-1   | 2024-09-13 07:19:00.388 UTC [30] WARNING:  no usable system locales were found
scheduler-1  | python: 1: /docker-entrypoint.sh: not found
api-1        | gunicorn: 1: /docker-entrypoint.sh: not found
postgres-1   | performing post-bootstrap initialization ... ok
```
what do you think the problem be?

raphodn:
Could you create issues for both of these subjects ? 🙏

PR #498, opened 2024-10-04
feat(Prices): new field to allow users to input the quantity bought

raphodn:
### What

See #499

New Price field (specific to RECEIPT): `receipt_quantity`

Add validation, and API setup (read & write)


open_prices/prices/models.py
```diff
@@ -361,6 +369,16 @@ def clean(self, *args, **kwargs):
                                     "proof",
                                     f"Proof {PROOF_FIELD} ({proof_field_value}) does not match the price {PROOF_FIELD} ({price_field_value})",
                                 )
+                if proof.type == proof_constants.TYPE_RECEIPT:
```
raphodn, line 372: what about GDPR proof types ?

raphodn: fixed here : https://github.com/openfoodfacts/open-prices/pull/498/commits/0e00b8aa1fbc6e336e8e9f2e6b4aba473b980464

open_prices/prices/models.py
```diff
@@ -133,6 +134,13 @@ class Price(models.Model):
         related_name="prices",
     )
 
+    receipt_quantity = models.PositiveBigIntegerField(
```
TTalex, line 137: I wouldn't expect this value to reach BigInteger levels.
`PositiveSmallIntegerField` is probably enough (<32767)

raphodn: nice catch ! I think I went a bit fast and copy/pasted the setup of the `osm_id` field. will change to `PositiveIntegerField`

PR #517, opened 2024-10-11
fix(Proofs): Deny users to delete a proof with prices

raphodn:
### What

Currently the API allows users to delete proofs even if the proof has prices.
(it would not delete prices, but set their proof_id to None)

To enforce data quality, and avoid mistakes, we restrict the deletion to only proofs without any prices.

What if the user linked the prices to a wrong proof for instance ? Will need to upload a new proof first, and the prices again to this new proof (maybe allow one day to update the price's proof, or update the proof image ?)

### Linked PRs

- https://github.com/openfoodfacts/open-prices/pull/197

### DIFF
--- a/open_prices/api/proofs/tests.py
+++ b/open_prices/api/proofs/tests.py
@@ -5,6 +5,8 @@ from django.urls import reverse
 from PIL import Image
 
 from open_prices.locations import constants as location_constants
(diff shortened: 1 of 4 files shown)


monsieurtanuki:
Just tested: "Cannot delete proof with associated prices" 👍

PR #529, opened 2024-10-19
fix(Tests): Remove proof images on test tearDown

raphodn:
### What

Currently the folder img/0001 keeps adding new images when running the tests (Proof API upload tests), but there is no mechanism to delete them.

"Hack" to fix this


raphodn:
actually it only fixes on "success" tests (where the Proof is actually created). 
does not work on 40X tests

PR #543, opened 2024-10-30
feat: run OCR extraction on every new image

### LINKED ISSUE #320: Proof: send to Google Cloud Vision, and store the matching JSON output
### Problem

- we should probably send all prices receipts and prices images to Google Cloud Vision, and store the matching JSON output

### Links

- https://cloud.google.com/vision/docs
- https://github.com/openfoodfacts/robotoff/blob/ef962195e66236a401356573595b0861186394e0/scripts/ocr/run_ocr.py#L32
- https://github.com/openfoodfacts/openfoodfacts-python/blob/develop/openfoodfacts/ocr.py#L1201 & https://github.com/openfoodfacts/openfoodfacts-python/blob/develop/openfoodfacts/ocr.py#L295
- https://github.com/clovaai/cord

### Why

- NSFW detection
- Siret Detection/Basic receipt data for simplistic pre-population
- Written barcode on shelf detection
- …
- detect the number of items, and make it available so that we can nudge the user properly
- detect the total
- detect date
- detect location (country at least)
- detect currency

raphael0202:
Fixes #320

### DIFF
--- a/open_prices/api/proofs/views.py
+++ b/open_prices/api/proofs/views.py
@@ -1,4 +1,6 @@
+from django.conf import settings
 from django_filters.rest_framework import DjangoFilterBackend
+from django_q.tasks import async_task
 from drf_spectacular.utils import extend_schema
 from rest_framework import filters, mixins, status, viewsets
 from rest_framework.decorators import action
@@ -75,6 +77,10 @@ class ProofViewSet(
                 status=status.HTTP_400_BAD_REQUEST,
             )
         file_path, mimetype, image_thumb_path = store_file(request.data.get("file"))
+        async_task(
+            "open_prices.proofs.utils.run_ocr_task",
+            f"{settings.IMAGES_DIR}/{file_path}",
+        )
         proof_create_data = {
             "file_path": file_path,
             "mimetype": mimetype,
--- a/open_prices/proofs/utils.py
+++ b/open_prices/proofs/utils.py
@@ -1,11 +1,20 @@
+import base64
+import gzip
+import json
+import logging
 import random
 import string
+import time
 from mimetypes import guess_extension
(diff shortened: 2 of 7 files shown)


open_prices/proofs/utils.py
```diff
@@ -124,3 +133,73 @@ def store_file(
...
+
+    data = run_ocr_on_image(image_path, api_key)
+
+    if data is None:
+        return
+
+    data["created_at"] = int(time.time())
+
+    with gzip.open(ocr_json_path, "wt") as f:
+        f.write(json.dumps(data))
```
raphodn, line 203: so it stores the result in a jsonl.gz file next to the image ?

raphodn: if we re-run on the same image, will it override ?

raphael0202: > so it stores the result in a jsonl.gz file next to the image ?
Yes exactly!

raphael0202: > if we re-run on the same image, will it override ?

It depends on the value of `override`

raphodn: ah ouiii bien vu !

raphodn:
todo : avoid running OCR extraction in the testruns + delete images

open_prices/api/proofs/views.py
```diff
@@ -75,6 +77,10 @@ def upload(self, request: Request) -> Response:
                 status=status.HTTP_400_BAD_REQUEST,
             )
         file_path, mimetype, image_thumb_path = store_file(request.data.get("file"))
+        async_task(
```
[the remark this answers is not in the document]
raphodn: done in #549

PR #545, opened 2024-11-04
chore: add CORS headers to images

raphael0202:
Otherwise, it can't be loaded from other website, ex: Label Studio instance

It was already tested (and hotfixed) in production.


raphodn:
@raphael0202 I have the same issue for data dumps. I created an issue #558

PR #548, opened 2024-11-06
feat: allow users to submit category_tag with language prefix

raphael0202:
Related to https://github.com/openfoodfacts/smooth-app/issues/5553

edit: as discussed in the smooth-app issue, I introduced a change when it comes to `category_tag`: we don't require anymore the `category_tag` to be valid with respect to the taxonomy.
The goal is to prevent silent failure from price addition, as it's done in a background task.
This is also more aligned with Open Food Facts philosophy to accept data as it is, and built upon it using taxonomies (or correct it in the moderation stage).


poetry.lock
raphodn, line 1: I updated the openfoodfacts-python package seperately in #550

raphodn: So you can delete the poetry changes :)

open_prices/prices/tests.py
```diff
@@ -108,11 +108,20 @@ def test_price_without_product_validation(self):
...
+                price_per=price_constants.PRICE_PER_KILOGRAM,
+            )
+        self.assertEqual(
+            cm.exception.messages[0],
+            "Invalid value: 'test', expected value to be in 'lang:tag' format",
+        )
+        PriceFactory(
             product_code=None,
-            category_tag="test",
+            category_tag="fr: Grenoble",  # valid (even if not in the taxonomy)
```
raphodn, line 124: @raphael0202 I added a test. it passes. should it ?

raphael0202: Yes it should! That's one of the major change, as long as the category is language prefixed properly, we accept it as a valid value.

raphodn (approved):

PR #557, opened 2024-11-09
feat(proofs): New endpoint to extract price data from images (with Gemini)

TTalex:
### What
- Added a new endpoint `POST /api/v1/proofs/process_with_gemini`
- This requires a new .env variable `GOOGLE_GEMINI_API_KEY`
- Also adds a new dependency `google-generativeai`
- The endpoint accepts a list of `files` that it sends to gemini with a parsing prompt
- Files should be images of labels
- The reply from gemini is sent directly to the frontend. It contains a list of objects, one for each file, check the `gemini.py` file for a list of expected values for `origin`, `product` and `unit` fields.
```
{
  "labels": [
    {
      "barcode": "3235223055203",
      "organic": false,
      "origin": "unknown",
      "price": 3.85,
      "product": "en:cherry-tomatoes",
      "unit": "KILOGRAM"
    },
    {
      "barcode": "335430001032",
      "organic": false,
      "origin": "unknown",
      "price": 2.19,
      "product": "en:strawberries",
      "unit": "KILOGRAM"
    }
  ]
}
```
### Part of 
- This relates to the Contribution Assistant experiment
- It is an adapted merge from https://github.com/TTalex/open-prices-contribution-tools

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -281,3 +281,8 @@ OFF_USER_AGENT = "open-prices/0.1.0"
 # ------------------------------------------------------------------------------
 
 GOOGLE_CLOUD_VISION_API_KEY = os.getenv("GOOGLE_CLOUD_VISION_API_KEY")
+
+# Google Gemini API
+# ------------------------------------------------------------------------------
+
+GOOGLE_GEMINI_API_KEY = os.getenv("GOOGLE_GEMINI_API_KEY")
--- a/open_prices/api/proofs/serializers.py
+++ b/open_prices/api/proofs/serializers.py
@@ -49,3 +49,11 @@ class ProofUpdateSerializer(serializers.ModelSerializer):
     class Meta:
         model = Proof
         fields = Proof.UPDATE_FIELDS
+
+class ProofProcessWithGeminiSerializer(serializers.ModelSerializer):
+    files = serializers.ListField(child=serializers.FileField(required=True, use_url=False))
+    mode = serializers.CharField() # TODO: this mode param should be used to select the prompt to execute, unimplemented for now
+    
+    class Meta:
+        model = Proof # TODO: this has nothing to do with the Proof model, fixme
+        fields = ["files", "mode"]
--- a/open_prices/api/proofs/serializers.py
+++ b/open_prices/api/proofs/serializers.py
@@ -51,14 +51,10 @@ class ProofUpdateSerializer(serializers.ModelSerializer):
         fields = Proof.UPDATE_FIELDS
 
 
-class ProofProcessWithGeminiSerializer(serializers.ModelSerializer):
+class ProofProcessWithGeminiSerializer(serializers.Serializer):
     files = serializers.ListField(
         child=serializers.FileField(required=True, use_url=False)
     )
     mode = (
         serializers.CharField()
     )  # TODO: this mode param should be used to select the prompt to execute, unimplemented for now
-
-    class Meta:
-        model = Proof  # TODO: this has nothing to do with the Proof model, fixme
-        fields = ["files", "mode"]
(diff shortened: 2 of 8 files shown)


config/settings.py
```diff
@@ -281,3 +281,8 @@
 # ------------------------------------------------------------------------------
 
 GOOGLE_CLOUD_VISION_API_KEY = os.getenv("GOOGLE_CLOUD_VISION_API_KEY")
+
+# Google Gemini API
+# ------------------------------------------------------------------------------
+
+GOOGLE_GEMINI_API_KEY = os.getenv("GOOGLE_GEMINI_API_KEY")
```
raphodn, line 288: @teolemon how do GOOGLE API works, could we re-use the GOOGLE_CLOUD_VISION_API_KEY ?
If not, could you add OFF's GOOGLE_GEMINI_API_KEY somewhere so that when we deploy in staging/prod the endpoint works ? 
Thanks !!

open_prices/api/proofs/serializers.py
```diff
@@ -49,3 +49,16 @@ class ProofUpdateSerializer(serializers.ModelSerializer):
     class Meta:
         model = Proof
         fields = Proof.UPDATE_FIELDS
+
+
+class ProofProcessWithGeminiSerializer(serializers.ModelSerializer):
```
raphodn, line 54: if nothing to do with the Proof serializer, you can simply rename to 
```suggestion
class ProofProcessWithGeminiSerializer(serializers.Serializer):
```

raphodn: https://www.django-rest-framework.org/api-guide/serializers/

raphodn: and remove the Meta

open_prices/api/proofs/views.py
```diff
@@ -94,3 +97,21 @@ def upload(self, request: Request) -> Response:
...
+
+    @extend_schema(request=ProofProcessWithGeminiSerializer)
+    @action(
+        detail=False,
+        methods=["POST"],
+        url_path="process_with_gemini",
+        parser_classes=[MultiPartParser],
+    )
+    def process_with_gemini(self, request: Request) -> Response:
+        if not request.data.get("files"):
```
raphodn, line 109: if already set as required in the Serializer, then its maybe not needed here ?

pyproject.toml
```diff
@@ -65,6 +65,7 @@ django-cors-headers = "^4.4.0"
 sentry-sdk = {extras = ["django"], version = "^2.13.0"}
 django-solo = "^2.3.0"
 pillow = "^10.4.0"
+google-generativeai = "^0.8.3"
```
raphodn, line 68: the `poetry.lock` file is missing (explains why the CI is failing), you need to run 
```
poetry add google-generativeai=0.8.3
```

config/settings.py
teolemon: Those are scoped, if we're using a specific key for Open Prices, we can extend the scope. Otherwise, @raphael0202 may have a key dedicated to Gen AI.

open_prices/api/proofs/views.py
TTalex: Without it, it seems that requests without the files param result in a 200.

I still removed it, since nothing seems to break without it anyway, it just replies with an empty list.

raphodn:
The video you did is really helpful ! 
cc @serpico

pyproject.toml
raphael0202: Or `poetry lock` to lock and updates `poetry.lock` file

raphael0202:
I generated a `GOOGLE_GEMINI_API_KEY` and added it as a secret, in staging and prod.
You need to add it to `container-deploy.yml` so that it's available in .env file: https://github.com/openfoodfacts/open-prices/blob/main/.github/workflows/container-deploy.yml#L129

And in docker-compose.yml, to pass the envvar to containers.

PR #601, opened 2024-12-05
feat: return proof predictions in `GET /api/v1/proof` route

raphael0202:
By doing so, the client can know all predictions related to the user proof.

### DIFF
--- a/open_prices/api/proofs/serializers.py
+++ b/open_prices/api/proofs/serializers.py
@@ -2,7 +2,21 @@ from rest_framework import serializers
 
 from open_prices.api.locations.serializers import LocationSerializer
 from open_prices.locations.models import Location
-from open_prices.proofs.models import Proof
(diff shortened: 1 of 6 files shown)


open_prices/api/proofs/serializers.py
```diff
@@ -18,6 +32,7 @@ class Meta:
 
 class ProofFullSerializer(ProofSerializer):
     location = LocationSerializer()
+    predictions = ProofPredictionSerializer(many=True, read_only=True)
```
raphodn, line 35: Not sure we should return the predictions when fetching a "list" of proofs. It could be returned only in the "retrieve" endpoint

raphael0202: Indeed, we could restrict it to the retrieve endpoint!

raphodn: done here: https://github.com/openfoodfacts/open-prices/pull/605

PR #603, opened 2024-12-05
feat: improve the permission strategy related to proof

raphael0202:
- allow anonymous proof upload (a new proof.anonymous field tracks which proof was uploaded by an anonymous user)
- allow to retrieve and list all proofs, irrespective of your authentication status
- allow users to add a price to a proof they don't own
- allow moderators to update/delete a proof

Tracking history for the proof table is meant to be made in a future PR


open_prices/prices/models.py
```diff
@@ -463,12 +463,6 @@ def clean(self, *args, **kwargs):
                 )
 
             if proof:
-                if proof.owner != self.owner:
```
raphodn, line 466: we should not allow users to add prices on proofs they do not own such as RECEIPTs or GDPR_REQUESTs or SHOP_IMPORTs

open_prices/api/proofs/views.py
```diff
@@ -99,8 +118,13 @@ def upload(self, request: Request) -> Response:
         serializer.is_valid(raise_exception=True)
         # get source
         source = get_source_from_request(self.request)
+        # Here, user is either a `open_prices.users.models.User` (if
+        # authenticated) or an `django.contrib.auth.models.AnonymousUser`
+        # (if not authenticated)
+        user = self.request.user
+        owner = None if user.is_anonymous else user.user_id
```
raphodn, line 125: None ? Shouldn't we set it to `openfoodfacts-contributors` instead ?

raphodn (commented):
Sorry but I think this PR should be split into multiple PRs, too many changes in one go it's hard to review and understand which line does what :sweat:

open_prices/api/proofs/views.py
raphael0202: After a bit of thought, it was easier to add a new flag (`anonymous`) instead of using a hardcoded owner.

open_prices/prices/models.py
raphael0202: True!

open_prices/users/models.py
```diff
@@ -33,9 +33,14 @@ class Meta:
         verbose_name = "User"
         verbose_name_plural = "Users"
 
+    @property
```
raphael0202, line 36: @raphodn This is important addition, there was a bug here, because we just checked that the `User` instance had a `is_authenticated` method when doing `if user.is_authenticated` (which is always true)

open_prices/api/proofs/views.py
raphodn: I'd rather mimick what is done in OFF.
Also avoids an extra field (with extra rules where is_anonymous can only be True if no owner is set, etc).

open_prices/users/models.py
raphodn: this requires a dedicated "fix" PR :pray: 
and to be tested, because it's probably here for a reason ? (that I can't remember) :thinking:

raphael0202: The method is here because we use it during proof upload. It was a hidden bug, as we used to block all non authenticated requests (but not anymore), so the user was always authenticated.

raphael0202:
I close this PR, as the only remaining feature (moderator edition) was implemented.

PR #613, opened 2024-12-09
fix: use openfoodfacts.ml module for proof classification

raphael0202:
I moved ML related code for object detection and image classification to `openfoodfacts` library, in a new `ml` module.
This module is used here in Open Prices, and will be used in Robotoff as well.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -24,23 +24,24 @@ files = [
 
 [[package]]
 name = "anyio"
-version = "4.4.0"
+version = "4.7.0"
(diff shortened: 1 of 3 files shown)


poetry.lock
raphodn, line 1: fyi you probably ran a `poetry update` instead of a `poetry update openfoodfacts` so all of the other packages were updated. 
I'd rather update the rest of the dependencies in a seperate PR to be sure there is no impact :pray:

PR #616, opened 2024-12-11
fix: fix bug in directory selection for proof image

raphael0202:
Solves #615

When the proof image folder reached the limit of 1000, we created a new image dir and stored the image in it, but without the 0 padding. This directory was never used again.
That's why we had directories named `1`, `2`,... `10` in the image dir.


raphael0202:
I've tested it locally, it works as expected.
I've cleaned the prod directory by manually moving proof images to their right locations.
I'm merging now, to prevent further cleaning!

PR #618, opened 2024-12-11
feat: listen to updates from Product Opener using Redis

raphael0202:
Just like Robotoff does.

### DIFF
--- a/docker/dev.yml
+++ b/docker/dev.yml
@@ -35,3 +35,20 @@ services:
   scheduler:
     <<: *api-base
     command: ["python", "manage.py", "qcluster"]
+
+  update-listener:
+    <<: *api-base
+    command: ["python", "manage.py", "run_update_listener"]
+
+  redis:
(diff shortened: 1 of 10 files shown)


docker/dev.yml
```diff
@@ -35,3 +35,20 @@ services:
   scheduler:
     <<: *api-base
     command: ["python", "manage.py", "qcluster"]
+
+  update-listener:
```
raphodn, line 39: 1) does this mean we can disable the cron tasks that daily sync the whole DB ? or should we have both running for a while ?
2) does redis send us changes from all flavors ?

raphael0202: > does this mean we can disable the cron tasks that daily sync the whole DB ? or should we have both running for a while ?
I prefer having both running for a while

> does redis send us changes from all flavors ?
It should, but here again we should make sure it works well in prod before disabling the daily dump import

Issue #611, opened 2024-12-08
Add a `price_tags` table to store predictions


raphael0202:
In #526, we've talked about how to speed up price addition using ML/AI (https://github.com/openfoodfacts/open-prices/discussions/526#discussioncomment-11323772).

For price tags, the workflow seems now clear:

1. user upload their proof
2. an object detection model identifies individual price tags. The object detection model returns the coordinates of each detected bounding box, along with a confidence score.
3. we extract information from the price tags: price, EAN or product category, price type, organic or not
4. we let the user (and later other users) fix and validate the prediction

Step (2) is currently in development. Once the proof is uploaded, we would ideally run all ML models as an async job.
For each detected price tag, we need to store some information:

- the bounding box of the price tag with respect to the original proof
- the detected price
- the detected category and/or EAN
- organic/non-organic
- the price per (per kg or per unit)
- origins (for raw products)

I suggest we create a new `price_tags` table. Storing intermediate data in a new table is necessary to allow performing async processing. The user won't have to wait for models to run, and we can distribute workload (=price validation) between contributors. Besides, it allows having gold truth data evaluating and training models that extract information from price tags.

### Schema

- `id`: ID of the price tag
- `proof_id`: FK to the proof
- `created`: creation datetime
- `updated`: datetime of last update
- `bounding_box`: the coordinates of the bounding box, in relative coordinated as (y_min, x_min, y_max, x_max). The origin is the top-left corner of the image.
- `status`: The status of the price tag: either a price is already linked for this price tag (status=1), it may be waiting for approval or completion (`status=null`), or it may be invalid (`status=0`, the information cannot be read or is hidden). Only `price_tag`s with `null` status will be suggested in a "Hunger Games-like" games.

Fields pre-filled by the extraction model (currently Gemini):
- `predicted_product_code`
- `predicted_category_tag`
- `predicted_price`
- `predicted_price_per`
- `predicted_price_is_discounted`
- `price_without_discount`
- `predicted_currency`
- `predicted_labels_tags`
- `predicted_origins_tags`

Fields that are derived from predictions, validated by the user. These fields are null before the price tag is validated by the user.

- `price`
- `price_is_discounted`
- `price_without_discount`
- `price_per`
- `product_code`
- `category_tag`
- `currency`
- `labels_tags`
- `origins_tags`
- `models_info` (JSONB): extra information about the version of the models that generated the prediction, about which model was used for what, etc. 

[...]

PR #628, opened 2024-12-16
feat: add `price_tag` table and routes

raphael0202:
Linked issue: #611

This PR only creates the `price_tag` table and allow users to create/update/delete new price tags.

A few remarks:
- price tags are not really deleted but flagged with `status=0`, so that we know for the images for which the object detector failed
- users can update the bounding box of an existing price tag, associate a price tag with a price, or update the status (including marking a price tag as deleted), but no more
- I didn't implement a full historisation of actions, but only the user who created the price tag (null if it was created by a model), and the user who last updated a price tag. It should be good enough for now.
- I had to create a new `open_prices.proofs.models` module to prevent circular imports 

Another PR will take care of implementing the creation of new price tags from the object detection model, and another one will add a new `price_tag_predictions` table and launch Gemini as worker tasks to populate this new table.


open_prices/proofs/models/price_tag.py
```diff
@@ -0,0 +1,150 @@
...
+    )
+    bounding_box = ArrayField(
+        base_field=models.FloatField(),
+        verbose_name="Coordinates of the bounding box, in the format [y_min, x_min, y_max, x_max]",
+    )
+    status = models.IntegerField(
+        choices=constants.PRICE_TAG_STATUS_CHOICES,
+        null=True,
+        blank=True,
+        verbose_name="The annotation status. Possible values are: "
```
raphodn, line 47: this info should be in the choices enum. the verbose_name should just be a short label :)

open_prices/proofs/migrations/0007_pricetag.py
```diff
@@ -0,0 +1,117 @@
...
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                (
+                    "created",
+                    models.DateTimeField(
+                        default=django.utils.timezone.now,
+                        verbose_name="When the tag was created in DB",
```
raphodn, line 32: use [help_text](https://docs.djangoproject.com/en/5.1/ref/models/fields/#help-text) for these kind of descriptions, and make the verbose_name short (or leave it empty)

open_prices/proofs/models/price_tag.py
raphodn, line 1: should we rather move this in a dedicated `ml` or `predictions` app ?
- avoids circular dependencies (though that can probably be avoid differently in any case)
- all the ML logic in one place

raphodn, line 1: Should we move this into a dedicated `ml` or `predictions` app ?
- to have all the ML logic in one single place
- to avoid recursive dependencies (though we could also avoid it in a different manner i believe)

open_prices/api/urls.py
```diff
@@ -23,6 +23,7 @@
 router.register(r"v1/products", ProductViewSet, basename="products")
 router.register(r"v1/proofs", ProofViewSet, basename="proofs")
 router.register(r"v1/prices", PriceViewSet, basename="prices")
+router.register(r"v1/price_tags", PriceTagViewSet, basename="price-tags")
```
raphodn, line 26:
```suggestion
router.register(r"v1/price-tags", PriceTagViewSet, basename="price-tags")
```

raphodn: recommended to use dash (hyphen) instead of underscore in URLs

raphodn: (yeah i know the gemini url doesn't match I'll take the time to refactor at somepoint :)

open_prices/proofs/models/price_tag.py
raphodn: In any case, I added a commit to revert your model separation :sweat_smile: https://github.com/openfoodfacts/open-prices/pull/628/commits/7b589dfc7c1e2f18b3e217a3763fbe8b9d867c58
- it fixed the recursive import
- it avoids unnecessary complexity :)

raphael0202: That's a good idea to create a `predictions` app. Let's do it later though!

open_prices/proofs/migrations/0007_pricetag.py
raphael0202: Indeed, I should have checked better Copilot suggestion ;)

PR #630, opened 2024-12-17
feat: save Gemini prediction in `price_tag_predictions` table

raphael0202:
We now save Gemini prediction in `price_tag_predictions` table.
I really struggled with unit-testing this method (as `unittest.mock.patch` function didn't work for some weird reason), so I leave it as-it for now.
I also moved Gemini code to `open_prices.proofs.ml` as it belongs here (and not in `open_prices.commons`), we will maybe split it again once we have a `predictions` django app.


raphodn (approved):
:rocket:

PR #631, opened 2024-12-17
feat: improve gemini processing

raphael0202:
- Update price tag prediction after a user updates a price tag
- create a price tag prediction after a user manually creates a price tag
- return price tag predictions when listing/retrieving price tags

### DIFF
--- a/open_prices/api/proofs/tests.py
+++ b/open_prices/api/proofs/tests.py
@@ -418,7 +418,8 @@ class PriceTagListApiTest(TestCase):
         # We only have 2 queries:
         # - 1 to count the number of price tags
         # - 1 to get the price tags and their associated proof
-        with self.assertNumQueries(2):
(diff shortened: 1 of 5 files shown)


open_prices/api/proofs/tests.py
```diff
@@ -418,7 +418,8 @@ def test_price_tag_list(self):
         # We only have 2 queries:
         # - 1 to count the number of price tags
         # - 1 to get the price tags and their associated proof
-        with self.assertNumQueries(2):
+        # - 1 to get the price tag predictions (prefetch related)
```
raphodn, line 421: we only have *3 queries

raphodn (approved):

open_prices/api/proofs/tests.py
raphodn: also occurences of "return full price" instead of "return full price_tag"

PR #648, opened 2024-12-22
fix: process one image at a time with Gemini

raphael0202:
Gemini sometimes skips some images when prediction price tag labels, leading to mismatch between price tag and predictions.
We process here one price tag at a time, preventing this from happening.

### DIFF
--- a/open_prices/api/proofs/views.py
+++ b/open_prices/api/proofs/views.py
@@ -25,7 +25,7 @@ from open_prices.api.proofs.serializers import (
 from open_prices.api.utils import get_source_from_request
 from open_prices.common.authentication import CustomAuthentication
 from open_prices.common.constants import PriceTagStatus
-from open_prices.proofs.ml import extract_from_price_tags
+from open_prices.proofs.ml import extract_from_price_tag
 from open_prices.proofs.models import PriceTag, Proof
 from open_prices.proofs.utils import store_file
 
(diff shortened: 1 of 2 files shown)


open_prices/api/proofs/views.py
```diff
@@ -126,8 +126,8 @@ def upload(self, request: Request) -> Response:
     def process_with_gemini(self, request: Request) -> Response:
         files = request.FILES.getlist("files")
         sample_files = [PIL.Image.open(file.file) for file in files]
-        res = extract_from_price_tags(sample_files)
-        return Response(res, status=status.HTTP_200_OK)
+        labels = [extract_from_price_tag(sample_file) for sample_file in sample_files]
+        return Response({"labels": labels}, status=status.HTTP_200_OK)
```
raphodn, line 130: the response changes ? so that we change the frontend integration

raphodn (approved):

open_prices/proofs/ml.py
```diff
@@ -161,8 +162,49 @@ class Labels(typing.TypedDict):
...
+            image,
+        ],
+        generation_config=genai.GenerationConfig(
+            response_mime_type="application/json", response_schema=Label
+        ),
+    )
+    return json.loads(response.text)
+
+
 def extract_from_price_tags(images: Image.Image) -> Labels:
```
raphodn, line 199: I re-added the previous function + added a warning: https://github.com/openfoodfacts/open-prices/pull/648/commits/d813e0d2c8e4616ce004c6e48bbc3503c28d96b8

open_prices/api/proofs/views.py
raphael0202: The response is exactly the same, no? The `Labels` TypedDict doesn't exist anymore

raphodn: nevermind, no changes :)

Issue #637, opened 2024-12-19
Automatically match new prices with their PriceTag prediction counterpart


raphodn:
### Story

When a proof is uploaded, we now automatically extract and create related PriceTags.
But the user might be adding prices with the regular workflows (aka not using the PriceTag predictions).

To avoid the risk of having duplicate prices, we need to hide them, by changing their status + linking them to the created price.

### How

- a script to manage the history
- a check after each new price creation

raphael0202:
To avoid duplicating the work in the new contribution assistant, I would suggest adding a new `unlinked_price_tags_count` column to the `proof` table, that represents for `PRICE_TAG` proofs the number of prices that are not linked to a price tag.

By doing this, we can filter in the price validator assistant only price tags associated to proofs for which we added prices through the price tag assistant only.

Detecting price duplicates on the backend seems like a waste of time for users to me, as they already checked and validated the price: we just need to link the price tag to an existing price, which may be done in another game.

raphodn:
I was thinking of an automated way, not a user-game way :)

raphodn:
`````
for proof in Proof.objects.filter(type="PRICE_TAG").prefetch_related("prices", "price_tags", "price_tags__predictions"):
    if proof.price_tags.count() == 0:
        continue
    elif proof.prices.count() == 0:
        continue
    else:
        for price_tag in proof.price_tags.all():
            if price_tag.price_id is not None:
                continue
            elif price_tag.predictions.count() == 0:
                continue
            else:
                price_tag_prediction_data = price_tag.predictions.first().data
                for price in proof.prices.all():
                    if price.price_tag_id is not None:
                        continue
                    elif price.type == "PRODUCT" and (price.data["product_code"] == price_tag_prediction_data["barcode"]) and (price.data["price"] == price_tag_prediction_data["price"]):
                        price_tag.price_id = price.id
                        price_tag.save()
                        break
                    elif price.type == "CATEGORY" and (price.data["category_tag"] == price_tag_prediction_data["category"]) and (price.data["price"] == price_tag_prediction_data["price"]):
                        price_tag.price_id = price.id
                        price_tag.save()
                        break
                    else:
                        continue
```

raphodn:
We also have the case of duplicate price tags (one created by the AI, another created by the user). It'll be fixed soon in the Contribution Assistant, but we'll need a script to match/cleanup/link the existing data

[...]

PR #650, opened 2024-12-22
feat(Price tags): script to match historical PriceTag with their prices

raphodn:
### What

Move the script into a reusable command

See https://github.com/openfoodfacts/open-prices/issues/637#issuecomment-2557890286

### DIFF
--- a/open_prices/proofs/management/commands/match_price_tags_with_existing_prices.py
+++ b/open_prices/proofs/management/commands/match_price_tags_with_existing_prices.py
@@ -1,0 +1,91 @@
+import sys
+from collections import Counter
+
+from django.core.management.base import BaseCommand
+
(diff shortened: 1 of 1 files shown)


open_prices/proofs/management/commands/match_price_tags_with_existing_prices.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+
+    help = "Match price tags with existing prices."
+
+    def handle(self, *args, **options) -> None:  # type: ignore
+        self.stdout.write("Stats before...")
+        stats()
+
+        self.stdout.write("Running...")
+        for proof in Proof.objects.has_type_price_tag().prefetch_related(
```
raphodn, line 39: there is probably an alternate way to loop on PriceTags instead of Proofs :thinking:

PR #676, opened 2025-01-12
feat(Prices): API: new filter on product category_tags

### LINKED ISSUE #673: Price API: Allow filter on product categories tags
### Problem
I would like to fetch prices matching a specific product category, for example "en:hazelnut-spreads".

Currently, the API only allows filtering on `category_tag` for products without a barcode (and thus no `product` relation)

### Proposed solution
Add a `product_categories_tags__contains` filter

TTalex:
### What
- Allows for filtering using `product__categories_tags__contains` on price APIs

### Fixes bug(s)
- Fixes #673

### DIFF
--- a/open_prices/api/prices/filters.py
+++ b/open_prices/api/prices/filters.py
@@ -35,6 +35,9 @@ class PriceFilter(django_filters.FilterSet):
     created__lte = django_filters.DateTimeFilter(
         field_name="created", lookup_expr="lte"
     )
+    product__categories_tags__contains = django_filters.CharFilter(
+        field_name="product__categories_tags", lookup_expr="icontains"
+    )
 
     class Meta:
         model = Price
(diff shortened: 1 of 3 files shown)


TTalex:
I have no idea what I'm doing test-wise, and I cannot run them locally for a few reasons (outdated operating systems..). So it's quite a surprise that they do not fail.

Help welcomed :)

raphodn:
> I have no idea what I'm doing test-wise, and I cannot run them locally for a few reasons (outdated operating systems..). So it's quite a surprise that they do not fail.

Haha, it works indeed :) thanks a lot !!
I pushed an extra commit to cleanup some related stuff.

open_prices/api/prices/filters.py
```diff
@@ -35,6 +35,9 @@ class PriceFilter(django_filters.FilterSet):
     created__lte = django_filters.DateTimeFilter(
         field_name="created", lookup_expr="lte"
     )
+    product__categories_tags__contains = django_filters.CharFilter(
```
raphodn, line 38: now that we know this works, it'll be easy to add extra product filters :eyes:

raphodn (approved):

PR #683, opened 2025-01-19
fix(Proofs): Run price tag prediction only on PRICE_TAG proofs

raphodn:
### What

When diving into our usage of ML, I realized that we run the `price_tag_detection` model even on non-Price tag proofs.
Nevertheless, we already had a check afterwards that would avoid creating the corresponding `PriceTag` objects for these kinds of proofs.

So moved up the check, and will cleanup the history once it is deployed

### Stats

```
Proof.objects.count()
// 17485

ProofPrediction.objects.count()
// 34956

ProofPrediction.objects.exclude(proof__type="PRICE_TAG").count()
// 6292

Counter(ProofPrediction.objects.exclude(proof__type="PRICE_TAG").values_list("model_name", flat=True))
// {'price_proof_classification': 3152, 'price_tag_detection': 3140}
```

### DIFF
--- a/open_prices/proofs/ml.py
+++ b/open_prices/proofs/ml.py
@@ -553,6 +553,9 @@ def run_and_save_price_tag_detection(
     :return: the ProofPrediction instance created, or None if the prediction
         already exists and overwrite is False
     """
+    if proof.type != proof_constants.TYPE_PRICE_TAG:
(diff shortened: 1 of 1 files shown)


raphodn:
Ran the following command to the cleanup post-deploy
```
ProofPrediction.objects.exclude(proof__type="PRICE_TAG").filter(model_name="price_tag_detection").delete()
// 3143 objects deleted
```

PR #734, opened 2025-02-25
fix: adding the blur_faces function

prasanna7codes:
### What
- 
- I created a blur_faces function which uses google cloud vision to blur the face of the uploader,
-  I installed google cloud vision by pip install google-cloud-vision

### Fixes bug(s)
- 

### Part of 
- 
- fixing the face issue, 
-

### DIFF
--- a/open_prices/proofs/utils.py
+++ b/open_prices/proofs/utils.py
@@ -5,9 +5,13 @@ from decimal import Decimal
 from mimetypes import guess_extension
 from pathlib import Path
 
+from google.cloud import vision
+import io
+
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


raphael0202:
Hello @prasanna7codes,
Sorry for not coming back to you sooner, I was in holidays. It's not really the way we want to implement it. First, we already have Google Cloud Vision results on the Open Prices server (ex: prices.openfoodfacts.org/img/0075/TlOVlNNMir.json.gz).
If we use Google Cloud Vision, we should add the faceAnnotation feature to the list of requested features, so that we can leverage it to detect and blur faces.
I'm closing this PR, as it requires additional discussion before implementation.

PR #744, opened 2025-03-04
feat(Users): new price count on owned (and not owned) proofs

raphodn:
### What

3 new user stats:
- price_in_proof_owned_count
- price_in_proof_not_owned_count
- price_not_owned_on_proof_owned_count


open_prices/users/models.py
```diff
@@ -13,6 +13,7 @@ class User(models.Model):
         "price_type_group_community_count",
         "price_type_group_consumption_count",
         "price_currency_count",
+        "price_from_proof_owned_count",
```
raphodn, line 16: not 100% satisfied with the field's name ^^

TTalex: I don't mind it, maybe replace from with in:  `price_in_proof_owned_count`

Issue #583, opened 2024-11-27
Proofs: allow users to specify that a receipt is not theirs


raphodn:
### Story

On the path to a user "consumption data dashboard", some proofs/prices are added even though they are not ours.

Allow users to flag/hide these prices so that they won't appear

serpico:
Would it be at the price level ( several prices on the receipt, some belonging to the user others not ) or at the proof level ?

raphodn:
The feedback by the user was at the proof level (adding prices bought by a friend who doesn't want to participate in OP)

serpico:
I asked because I got the case this morning, 1 receipt 2 products, one was not mine. I can't really picture how powerful/versatile the "consumption data dashboard" will be, but on principle, I don't wanna miss out :-)

raphodn:
Can't figure out a good wording for a field that could be added in both the `Proof` and `Price` models.
- Proof.receipt_not_mine & Price.not_mine ("mine" is too familiar)
- Proof.exclude_from_consumption & Price.exclude_from_consumption (exclude and "not mine" is different 🤔 )
- Proof.not_owned_by_user & Price.not_owned_by_user
- ?

raphodn:
Indeed having `owner` in the field name is a good idea.

For positive (belong) or negative (not/exclude) wording, I need some more thoughts, if the field should have some True or False default value, or if it should be Null most of the time 🤔

raphodn:
An alternative could be to use this field to help determine if the proof/prices are "consumption" or not (the rest being community - see #727)
For instance a `owner_consumption`, that would be False by default, and set to True in the UI.
The main hurdle is receipt proofs coming from other sources (like the API, or the Mobile), where this field will likely not be implemented anytime soon, and so proofs will be set to False. We could set them to True ? Or set the default value to None 🤔

serpico:
> An alternative could be to use this field to help determine if the proof/prices are "consumption" or not (the rest being community - see [#727](https://github.com/openfoodfacts/open-prices/issues/727))
>

If there is really a set of features down the road that allows to sum up each and every product bought and added to OP by one user it could be even used to try to track food intake ( let's say my OP account shows that I bought 10 or 11kg of sugar over a period of 12 months, and I know for a fact that I'm the only one eating that sugar [short of 1 full 1kg in stock and 1kg currently in use]. 

[...]

PR #750, opened 2025-03-06
feat(Proofs): new owner_consumption field

raphodn:
### What

See #583

- new `Proof.owner_consumption` boolean field
- default to None
- can only be set (True or False) if the proof is a RECEIPT or a GDPR_REQUEST (consumption group)
- in the migration, init to True for all existing consumption proofs


TTalex (approved):
Seems fine for a V1.

But we'll probably have to think about it on a price level at some point, if users want to divide the proof between multiple people, such as family members.

PR #753, opened 2025-03-08
feat(Prices): API: new filter on proof owner_consumption

raphodn:
### What

Similar to #741, and following #750
Add a new API price filter on `proof__owner_consumption`

### DIFF
--- a/open_prices/api/prices/filters.py
+++ b/open_prices/api/prices/filters.py
@@ -31,6 +31,9 @@ class PriceFilter(django_filters.FilterSet):
         field_name="proof__type",
         choices=proof_constants.TYPE_CHOICES,
     )
+    proof__owner_consumption = django_filters.BooleanFilter(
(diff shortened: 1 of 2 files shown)


open_prices/api/prices/filters.py
```diff
@@ -31,6 +31,9 @@ class PriceFilter(django_filters.FilterSet):
         field_name="proof__type",
         choices=proof_constants.TYPE_CHOICES,
     )
+    proof__owner_consumption = django_filters.BooleanFilter(
```
raphodn, line 34: actually I don't know if it's the right strategy...
because it'll be tricky to filter for the opposite ("community" prices).
maybe instead create a new custom filter - for instance `proof__group` that could be "consumption" or "community", and we would manage the custom filtering somewhere in the backend

raphodn:
Closing in favor of https://github.com/openfoodfacts/open-prices/pull/776

PR #758, opened 2025-03-16
feat(Challenges): basic model and API endpoint

### LINKED ISSUE #691: New table to store challenge configuration
### What

After launching the first challenge in the frontend, and if we wish to launch more challenges in the future, then we need to start storing the data in the backend.

### Related issues & discussions

- https://github.com/openfoodfacts/open-prices-frontend/issues/934
- https://github.com/openfoodfacts/open-prices/discussions/509

TTalex:
### What

- A basic model to store Challenges information.
- API endpoint `/challenges` (GET only - list & retrieve)
- API doesn't include creation, update or deletion, since these actions should probably be restricted to admins.
- Model should probably be extended to hold end of challenge information, such as stats and rankings, as described in #691. However, this is not required for now.

### Screenshot

image

### Fixes bug(s)
- Fixes #691


raphodn (approved):
thanks :pray: :chestnut:

open_prices/challenges/models.py
```diff
@@ -0,0 +1,81 @@
...
+from open_prices.challenges import constants as challenge_constants
+from open_prices.common import utils
+
+
+class ChallengeQuerySet(models.QuerySet):
+    def published(self):
+        return self.filter(is_published=True)
+
+
+class Challenge(models.Model):
```
raphodn, line 15: did a couple of changes
- removed the `challenge_id` field (there is already an auto-generated `id` field)
- removed the `state` field and added a `status` property
- added a `is_published` boolean field
- added some validation on save (start & end dates, some fields required if published)

Issue #764, opened 2025-03-21
Product price estimate (average ? or last ? or range ?)


raphodn:
Similar to #124 & #254

The idea would be to create a new Product field to store an "average" price.
That could be easily displayed in the frontend (without having to do extra API calls)

### Why

If we want to display in the frontend some basic "price comparison" when searching or adding a product, we need to have a quick way to get an estimate price of the given product.

### Ideas

- average price ; mean, median, mode... 
- or last ?
- or a range ?
- calculation should depend on the amount of prices the product has
- filter on the last 3 months
- currency : what happens if the product has prices in multiple currencies ?
- filter out discounted prices
- document the calculation somewhere and make it easily accessible
- ?

raphodn:
> We could also use the [interquartile range](https://en.wikipedia.org/wiki/Interquartile_range) to detect if a price has a typo or is some type of outlier.
We would need a certain amount of prices per product to have this method effective.

@baslia [here](https://github.com/openfoodfacts/open-prices/discussions/319#discussioncomment-12039995)

raphodn:
How to name this field ?

- `price_indicative`
  - "Prix indicatif" in French has a (strict ?) definition of "Prix de vente conseillé" ([source](https://www.sciences-commerciales.com/index.php/Prix_indicatif))
- `price_agg`
- `price_op`

raphodn:
It could also be in a dedicated model. If we wanted to add multiple fields... Or a JSONField ?

baslia:
I would rather choose a simple outlier detection technique, as I am not sure we will have enough data to properly fit a model.
IQR would be a solid fondation.

raphodn:
I added a "Why" section in the first comment

> If we want to display in the frontend some basic "price comparison" when searching or adding a product, we need to have a quick way to get an estimate price of the given product.

To explain the idea behind this "price estimate", which is probably different from a more complex "price analysis"

PR #765, opened 2025-03-21
feat(Products): new utils to calculate the price estimate

raphodn:
- Linked to #764

### What

First step to calculate a "price estimate" for each Product in our DB


TaciteOFF:
This is interesting but calculate_price_estimate() doesn't filter by currency, as talked in https://github.com/openfoodfacts/open-prices/issues/764

PR #772, opened 2025-03-23
refactor(Stats): rename price stats per type (product & category)

raphodn:
### What

These 2 stats were created in #496.
Following a recent PR - #771 - we rename them to match the field name with the actual field value (and use the new querysets)
- price_type_product_code_count > price_type_product_count
- price_type_category_tag_count > price_type_category_count


open_prices/stats/models.py
```diff
@@ -70,8 +70,8 @@ class TotalStats(SingletonModel):
     )
 
     price_count = models.PositiveIntegerField(default=0)
-    price_type_product_code_count = models.PositiveIntegerField(default=0)
-    price_type_category_tag_count = models.PositiveIntegerField(default=0)
+    price_type_product_count = models.PositiveIntegerField(default=0)
```
raphodn, line 73: impact on mobile/dart integration ?

raphodn: (will need a web frontend PR in any case)

Freso: I’m not sure how Django models work, but could it be possible to add something like
```python
@property
def price_type_product_code_count(self) -> int:
    return self.price_type_product_count

@property
def price_type_category_tag_count(self) -> int:
    return self.price_type_category_count
```
to provide backwards compatibility until whatever "external" things using these have been updated?

PR #778, opened 2025-03-23
feat(ProofMl): Added ReceiptItem model and API

### LINKED ISSUE #681: Proof: use GenAI to extract receipt data (instead of basic OCR)
### Story

We already use GenAI (Google Gemini AI) to extract price tag data (see #569)
And we run OCR (Google Cloud Vision) to extract text (see https://github.com/openfoodfacts/open-prices/issues/320), but the data is stored as `json.gz` and is not used (nor is very easy to leverage, see a full example [here](https://github.com/openfoodfacts/open-prices/wiki/Proofs-x-AI))

What about trying to send receipt images to GenAI. With the correct prompt, we could probably get some interesting data out of it !

### Related issues & discussions

- #569
- https://github.com/openfoodfacts/open-prices/wiki/Proofs-x-AI

cc @TTalex & @odin-h

TTalex:
### What
- Proposed implementation of ReceiptItem model specified in #681
- Small change compared to spec: renamed `data` field to `predicted_data`
- Tests are yet to be written

### Fixes bug(s)
- Fixes #681


open_prices/proofs/ml.py
```diff
@@ -558,6 +558,33 @@ def create_price_tags_from_proof_prediction(
...
+        )
+        return []
+
+    created = []
+    for predicted_item in proof_prediction.data["items"]:
+        receipt_item = ReceiptItem.objects.create(
+            proof=proof,
+            proof_prediction=proof_prediction,
+            price=None,
+            order=len(created) + 1,
```
raphodn, line 580: rather use the index of the for loop (seems more logical).
can use `enumerate` to get it :)

open_prices/proofs/models.py
```diff
@@ -702,3 +702,57 @@ def price_tag_prediction_post_create_increment_counts(
...
+        on_delete=models.SET_NULL,
+        related_name="receipt_items",
+        null=True,
+        blank=True,
+        help_text="The price linked to this receipt item",
+    )
+    order = models.IntegerField(
+        help_text="The order of the item in the receipt. Item on top is 1.",
+    )
+    predicted_data = models.JSONField(
```
raphodn, line 736: `data` is used elsewhere. I prefer `predicted_data` :+1:

raphodn (approved):
go go go !

PR #784, opened 2025-03-28
chore(docs): create dotenv example file, ignore the local one

grunskis:
### What
`INSTALL.md` tells developers to create a local copy of the `.env.example` file, but this file doesn't exist. Instead `.env` exists which can accidentally be committed. To prevent that I created `.env.example` and added `.env` to the list of ignored files.

### Screenshot
None

### Fixes bug(s)
None

### Part of 
None


raphodn (approved):
thanks !

.gitignore
```diff
@@ -109,6 +109,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+.env
```
raphodn, line 112: I had to revert your PR in https://github.com/openfoodfacts/open-prices/pull/787

raphodn: the file is used in the Makefile

raphodn: the CI was raising errors...
```
api Pulling 
api Error manifest unknown
Error response from daemon: manifest unknown
```

raphodn: not sure how to fix this in the long run though..

raphodn: fyi in my local environment, I run the app without docker, passing in the env variables as prefixes in the terminal
```
POSTGRES_DB=open_prices POSTGRES_USER=... POSTGRES_PASSWORD=... POSTGRES_HOST=... ALLOWED_HOSTS=localhost,127.0.0.1 CSRF_TRUSTED_ORIGINS=http://127.0.0.1:8000 SECRET_KEY=... OAUTH2_SERVER_URL=https://world.openfoodfacts.org/cgi/auth.pl DEBUG=True poetry run python manage.py runserver
```

PR #814, opened 2025-04-27
refactor(Prices): new has_tag queryset

raphodn:
### What

Following #807

- new `Price.objects.has_tag(<tag:str>)` queryset
- add a test


open_prices/prices/tests.py
```diff
@@ -132,6 +133,11 @@ def test_calculate_stats(self):
             },
         )
 
+    def has_tags(self):
```
raphodn, line 136: typo, fixed in https://github.com/openfoodfacts/open-prices/pull/817

PR #832, opened 2025-05-08
fix(Prices): also look at category_tag for in_challenge checks

TTalex:
### What

Following #819

- Category prices are not included in challenges stats, even when their category matches.


raphodn (approved):
thanks :)
I added 2 commits to improve the queries

Issue #688, opened 2025-01-26
Price tags: levels of difficulty (ex: AI found an existing product or category)


raphodn:
### Story

Currently, all the Price Tags (with status=None) are visible in the Price Validation Assistant (web frontend).
But some PT are very hard to decipher, or simply not well detected. And we have new users who find the PVA too complex.

In order to build a simplified PVA, we need to isolate simple PTs from difficult ones.

### How

An idea is to sort them by magnitudes of difficulties : 
- level 1 : found a barcode corresponding to an existing product
- level 2 : found an existing category
- level 3 : found a valid barcode, but no product exists yet
- level 4 : the rest 😅 

We could add a field `difficulty_level` that would be calculated similarly to #656

The tricky bit is that this field would be stored on the prediction. So we then need to be able to filter the API on the predictions...

### Todo

- [x] first list of rules
- [x] new `PriceTag.tags` field to store the results: #806
- [x] new `PriceTagPrediction` helper methods to tell if the predicted barcode/product/category_tag is valid or not: #838
- [x] new signal to set the `PriceTag` tags after its `PriceTagPrediction` creation: #843
- [ ] run a script on the the `PriceTag` history

raphodn:
With the help of #806 (new `PriceTag.tags` field), we can label the PriceTags with the above rules

|Rule|tag label|
|-|-|
|prediction found a barcode corresponding to an existing product|`prediction-found-existing-product`|
|prediction found a valid barcode|`prediction-found-valid-barcode`|
|prediction found a category|`prediction-found-category`|

@TTalex sounds good ? ideas of extra rules, or better naming ?

TTalex:
Looks good !

raphodn:
ok I merged 2 PRs. and updated my comments above!

raphodn:
todo
- [x] refactoring : move the logic to ml.py
- [x] manage changes due to schema v2
    - following #885
- [x] manage specific barcode rules (e.g. Carrefour) currently done in the frontend
    - dedicated issue : https://github.com/openfoodfacts/open-prices/issues/1020
- [ ] run a script on the history (and the changes post-v2)
- [x] add a filter in the frontend UI

raphodn:
I'm a bit stuck on the shop-specific rules.
I'd like to implement them in the backend, but I don't want to change the output of Gemini nevertheless.
[...]

PR #838, opened 2025-05-09
feat(Price Tag): predictions: helper methods to know if the detected data is "valid"

raphodn:
### What

Following #834 (new `PriceTag.tags` field)
Linked to #688

### Todo in future PRs

- method to set a tag depending on "rules"
- signal to automatically run the method on PriceTagPrediction creation

### DIFF
--- a/open_prices/proofs/models.py
+++ b/open_prices/proofs/models.py
@@ -1,5 +1,6 @@
 import decimal
 
+import openfoodfacts
 from django.conf import settings
 from django.contrib.postgres.fields import ArrayField
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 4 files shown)


open_prices/proofs/models.py
```diff
@@ -759,6 +760,29 @@ class Meta:
...
+
+        if self.has_predicted_barcode_valid():
+            barcode = self.data.get("barcode")
+            if Product.objects.filter(code=barcode).exists():
+                return True
+        return False
+
+    def has_predicted_product_valid(self):
+        if self.data.get("product"):
+            category_tag = self.data.get("product")
```
raphodn, line 781: @TTalex slightly hard to juggle between "product" & "category" here in these predictions :sweat_smile:

TTalex: Yeah, naming isn't great :/
It should really be called something like "detected_category_tag" in the ml model ... but it's a bit late

raphodn: How late is it ? What happens if we rename the prompt to Gemini + script to update the history + changes in the Frontend. Now or never, no ? 🤪

TTalex: Yeah, good point.
The names fed to the prompt might be important for Gemini, but we could map them to proper names at the end of the process.

raphodn: i added some quick "category_tag" comments

PR #867, opened 2025-07-03
chore: added server block to the spec file generation

JagjeevanAK:
### What
added server block to the spec file generation to support the upcoming api doc generation framework

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -194,6 +194,61 @@ REST_FRAMEWORK = {
     "DEFAULT_PERMISSION_CLASSES": [],
 }
 
+# Build server URLs dynamically for OpenAPI documentation
+def get_api_servers():
+    """
+    Generate server configuration for OpenAPI schema based on environment.
+    This ensures that Fumadocs and other API documentation tools have proper
+    server endpoints for testing the API.
+    """
+    servers = []
+    
+    # Get the API port configuration
+    api_port = os.getenv("API_PORT", "127.0.0.1:8000")
+    
+    if DEBUG:
+        # Development servers
+        if ":" in api_port:
+            host, port = api_port.split(":")
+            dev_url = f"http://{host}:{port}/api"
+        else:
+            dev_url = f"http://localhost:{api_port}/api"
+        
+        servers.append({
+            "url": dev_url,
+            "description": "Development server"
+        })
+        
+        # Also add localhost variant for Docker development
+        servers.append({
+            "url": "http://localhost:8000/api",
+            "description": "Local Docker development server"
+        })
+    else:
+        # Production server
+        production_url = os.getenv("PRODUCTION_API_URL")
+        if production_url:
+            servers.append({
+                "url": production_url,
+                "description": "Production server"
+            })
+        else:
+            # Default production URL for Open Food Facts
+            servers.append({
+                "url": "https://prices.openfoodfacts.org/api",
+                "description": "Production server"
+            })
+    
+    # Add staging server if specified
+    staging_url = os.getenv("STAGING_API_URL")
+    if staging_url:
+        servers.append({
+            "url": staging_url,
+            "description": "Staging server"
+        })
+    
+    return servers
+
 SPECTACULAR_SETTINGS = {
     "TITLE": "Open Food Facts open-prices REST API",
     "DESCRIPTION": "Open Prices API allows you to add product prices",
@@ -210,6 +265,7 @@ SPECTACULAR_SETTINGS = {
     "ENUM_NAME_OVERRIDES": {
         "LocationOsmTypeEnum": "open_prices.locations.constants.OSM_TYPE_CHOICES"
     },
+    "SERVERS": get_api_servers(),
 }
 
 


config/settings.py
```diff
@@ -194,6 +194,58 @@
     "DEFAULT_PERMISSION_CLASSES": [],
 }
 
+
+# Build server URLs dynamically for OpenAPI documentation
+def get_api_servers():
```
raphael0202: Adding return type would be nice here

config/settings.py
```diff
@@ -194,6 +194,58 @@
...
+        # Also add localhost variant for Docker development
+        servers.append(
+            {
+                "url": "http://localhost:8000/api",
+                "description": "Local Docker development server",
+            }
+        )
+    else:
+        # Production server
+        production_url = os.getenv("PRODUCTION_API_URL")
```
raphael0202: The URL of the production server is not going to change, there is no need to use an envvar here.
You can your code by directly adding `https://prices.openfoodfacts.org/api`.

config/settings.py
```diff
@@ -194,6 +194,58 @@
...
+            # Default production URL for Open Food Facts
+            servers.append(
+                {
+                    "url": "https://prices.openfoodfacts.org/api",
+                    "description": "Production server",
+                }
+            )
+
+    # Add staging server if specified
+    staging_url = os.getenv("STAGING_API_URL")
```
raphael0202: Same here, the staging URL is `https://prices.openfoodfacts.net/api`, you can directly add it without having to look for an envvar.

raphael0202 (changes requested):

JagjeevanAK:
@raphael0202 Thank you for your review I will add the changes to it.

config/settings.py
```diff
@@ -194,6 +194,58 @@
...
+        servers.append(
+            {
+                "url": "http://localhost:8000/api",
+                "description": "Local Docker development server",
+            }
+        )
+    else:
+        # Production server
+        production_url = "https://prices.openfoodfacts.org/api"
+        if production_url:
```
raphael0202: There is no need for this if condition here, the variable is defined above.

config/settings.py
```diff
@@ -194,6 +194,58 @@
...
+            servers.append(
+                {
+                    "url": "https://prices.openfoodfacts.org/api",
+                    "description": "Production server",
+                }
+            )
+
+    # Add staging server if specified
+    staging_url = "https://prices.openfoodfacts.net/api"
+    if staging_url:
```
raphael0202: Same remark, There is no need for this if condition here.

.env.example
```diff
@@ -0,0 +1,60 @@
+# unify separator with windows style
```
raphael0202: For the moment we don't use `.env.example` but directly .env, can you remove this file from the PR?

.gitignore
```diff
@@ -145,3 +145,4 @@ www/static/admin
 www/static/django_extensions
 www/static/rest_framework
 gh_pages
+.env
```
raphael0202: Can you remove this line?

.env.example
JagjeevanAK: I agree but it is good practice to have an .env.example correct ?

config/settings.py
JagjeevanAK: ya you are correct my bad.

config/settings.py
JagjeevanAK: sorry for that will update the change

JagjeevanAK: In hurry I forgot that

.env.example
raphael0202: At Open Food Facts we use .env directly, without `.env.example`. It's something we can discuss, but even if we do so, this should anyway be done in a separate PR anyway.

JagjeevanAK:
hey @raphael0202 done 👍.

raphael0202 (approved):

config/settings.py
```diff
@@ -194,6 +194,47 @@
...
+        servers.append(
+            {
+                "url": "http://localhost:8000/api",
+                "description": "Local Docker development server",
+            }
+        )
+    else:
+        servers.append(
+            {
+                "url": "https://prices.openfoodfacts.org/api",
```
raphodn, line 230: if there is a new API_PORT env above, shouldn't we use it instead of hardcoding the URL here ?

raphodn: for instance the pre-prod environment ? https://prices.openfoodfacts.net

JagjeevanAK: Hey @raphodn, I already done that but @raphael0202 said we don't change the url that much so it is good to have an hardcoded one
Screenshot 2025-07-08 at 10 27 57 PM

JagjeevanAK: so you mean to have like this ?

```py
os.environ.get("API_PORT", "8000")
```

raphael0202: >Hey @raphodn, I already done that but @raphael0202 said we don't change the url that much so it is good to have an hardcoded one
Sorry, I meant prices.openfoodfacts.net in my comment, not prices.openfoodfacts.org

raphodn: - in the codebase there is already a reference to "API_PORT", and it seems to be a PORT, not a URL like above (127.0.0.1)
- adding a new env variable such as "HOST" (instead of API_PORT) seems like the way to go
- and modify containter-deploy.yml and set the .net & .org values there

PR #870, opened 2025-07-04
fix: fix bug in Redis update listener

raphael0202:
The update was failing for Open Beauty Facts.


open_prices/products/management/commands/run_update_listener.py
```diff
@@ -20,7 +20,7 @@ def process_redis_update(self, redis_update: RedisUpdate):
         if redis_update.product_type == "food":
             flavor = Flavor.off
         elif redis_update.product_type == "beauty":
-            redis_update.obf
+            flavor = Flavor.obf
```
raphodn, line 23: This means that Redis sends create & update info for all flavors, not just food ?

raphodn: If yes, then we can disable daily syncs for non-food flavors :)

PR #879, opened 2025-07-19
refactor(Price tags): Add new statuses 'no_barcode' (5) & 'other' (10)

TTalex:
Similar to #702

### What

Simply add a generic 'other' status, for price tags not matching existing one. Main usage would be for proofs with already added prices.

We also add a `no_barcode` status.


open_prices/proofs/constants.py
```diff
@@ -55,6 +55,7 @@ class PriceTagStatus(enum.IntEnum):
     not_readable = 2
     truncated = 3
     not_price_tag = 4
+    other = 5
```
raphodn: > Main usage would be for proofs with already added prices.

I'm not sure I understand
- if it's the main usage, why not be more explicit ? (instead of "other")
- what's the issue with "proofs with already added prices" ? even if the proof has prices added seperately, we could run the `match_price_tags_with_existing_prices` script more frequently instead ?

TTalex: > if it's the main usage, why not be more explicit ? (instead of "other")

I remember "other" being mentioned during last weekly call, for all cases not matching the existing ones. Another case could be when the barcode isn't present on the image, setting "unreadable" isn't correct.

I imagine a default choice can also be useful for confused users.

> what's the issue with "proofs with already added prices" ?

If I see that the price was already added, I would rather discard the price tag right away, instead of struggling to figure out a barcode.

This is probably a non-issue if mobile no longer pushes everything to the validation assistant, so we could also wait for that to get fixed.

> even if the proof has prices added seperately, we could run the match_price_tags_with_existing_prices script more frequently instead ?

Oh yeah, that helps, I forgot about that. Only helps if users aren't too quick, and price detection goes well.
How often does it currently run ? Weekly ?

TTalex: Now that I think about it, I seem to remember some cases where `match_price_tags_with_existing_prices` should have done something.

Example here: https://prices.openfoodfacts.org/proofs/33756

Proof was created on 2025-05-27, first price was created the same day using mobile, a price tag object was also created the same day, with a prediction matching on price (2.29).

5 weeks later, the price tag object is still in the PVA, and a duplicate is created.

raphael0202: I would definitely add a "no barcode" option.

raphodn: I agree we can start with adding a `no_barcode` option :)

And I need to improve the matching script 👍

raphodn (approved):

open_prices/proofs/constants.py
raphodn: fyi i added both a `no_barcode` (5) and a `other` (10) option.
but in the frontend we won't necessarily show the other option right now :) 
and with the other Raphael we thought about improving the PVA : wrote them in the corresponding issue : https://github.com/openfoodfacts/open-prices-frontend/issues/1137

PR #885, opened 2025-07-23
feat: improve price tag extraction using Gemini

raphael0202:
This PR improves both the speed and accuracy of extraction from price tags.
The speed is improved by sending requests to Gemini in parallel, using an event loop (asyncio).

The performance of the extraction is improved by:
- providing a JSON Schema with detailed descriptions and rules to follow, through Pydantic, to Gemini model
- enabling model reasoning during extraction

I've added the following modifications to the original schema:

- I added a new `type` (`PRODUCT` or `CATEGORY`) to know the price type (same as on Open Prices `price` model)
- If `type=PRODUCT`, wet set `category` to a new value `NO_CATEGORY` and `origin` to a new value `NO_ORIGIN`
- we don't have a `price` field anymore, but a list of `prices`. This allows to extract price values for more complex price tags. Example: we have 2 prices on the price tag, one for fidelity card members, one for the general public. It's also useful to extract the price per kg/l (as often displayed on price tags), to be able to compare it to the theoretical price per kg/l, using OFF data. Finally, it allows us to suggest in the front-end other detected prices, in case the model didn't extract the right value. Another improvement is that for `type=CATEGORY`, if the price is given for 500 g or 100g, we ask the model to convert it to the price per kg (it works well in my tests).
- I added the `blurriness`, `truncated` and `is_price_tag` fields. Adding new fields is cheap, and this data will be useful as a pre-categorization when training specific models.

This new schema obviously breaks the currently used schema, which means we will need to update the front-end as well.
I think we should version the schema of the extraction, so that the front-end doesn't break when we have multiple schema in DB.

Missing from this PR:
- support for VAT-less prices. In Japan, for example, most of the time 2 prices are displayed: with VAT and without.
- creating a new `price` field during Pydantic model export that is the most likely price, with support for some forms of discounted prices (ex sales), just like Open Prices does.

### DIFF
--- a/open_prices/api/proofs/views.py
+++ b/open_prices/api/proofs/views.py
@@ -138,7 +138,10 @@ class ProofViewSet(
     def process_with_gemini(self, request: Request) -> Response:
         files = request.FILES.getlist("files")
         sample_files = [PIL.Image.open(file.file) for file in files]
-        labels = [extract_from_price_tag(sample_file) for sample_file in sample_files]
+        labels = [
+            extract_from_price_tag(sample_file).parsed.model_dump()
+            for sample_file in sample_files
+        ]
         return Response({"labels": labels}, status=status.HTTP_200_OK)
 
 
--- a/open_prices/common/google.py
+++ b/open_prices/common/google.py
@@ -58,7 +58,20 @@ def get_genai_client() -> genai.Client:
     return genai.Client()
 
 
-def get_generation_config(response_schema) -> types.GenerateContentConfig:
+def get_generation_config(
+    response_schema: type, thinking_budget: int = -1
+) -> types.GenerateContentConfig:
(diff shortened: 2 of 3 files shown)


raphodn (approved):

open_prices/proofs/migrations/0021_pricetagprediction_schema_version_and_more.py
```diff
@@ -0,0 +1,31 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("proofs", "0020_alter_pricetag_status"),
+    ]
+
+    operations = [
+        migrations.AddField(
```
raphodn, line 12: we'll need to init to "1.0" for past predictions (no need to change the code, we can run a script instead)

Issue #620, opened 2024-12-11
Product sync: use OFF's Redis stream to get live updates


raphodn:
### Story

We currently sync all of OFF's product database every day (see https://github.com/openfoodfacts/open-prices/discussions/460 for more details).
But this system has some issues : 
- the sync can take over an 1h to run. every day
- if a user makes changes to a product (or simply adds a new one), it will take sometimes 24h to get the result/feedback in OP

OFF has a stream where every product creation & modification is available. We could plug OP on this stream.

raphodn:
Since #618 we now get live updates ! 

But only for OFF products I believe, not OBF nor OPF nor OPFF.

raphodn:
I see some improvements : 
- disable OFF DB daily sync now that we have redis
- we don't use the `diffs`  that Redis sends us, but call the OFF API instead
- do we still need to call `product_post_create_fetch_and_save_data_from_openfoodfacts` each time a product is created in OP ? if it's already unknown in OP, then it won't be available in OFF (as we fetch only OFF data, not OxF)

raphodn:
Closing. remaining stuff in #745

PR #887, opened 2025-07-24
refactor(Products): disable OxF sync now that we have Redis for all

raphodn:
### What

Following #870. Linked to #620


raphael0202:
Before disabling daily sync, we still need to support product type change: when a user change from OFF to OBF for example, we don't perform the switch in Open Prices DB.
See https://github.com/openfoodfacts/openfoodfacts-python/blob/develop/openfoodfacts/redis.py#L53 in the Open Food Facts repo to see what the event of a type switch looks like.

PR #888, opened 2025-07-25
feat: don't rely anymore on enums for Price tag validation

raphael0202:
(DO NOT MERGE)

Previously, we were using enum to ask Gemini to choose between specific origins and categories (for raw products).
The drawback of this approach is that:

- we have to maintain an enum that is up-to-date with respect to the origins and the categories.
- it increases the number of input tokens with send to Gemini, as the full enum in sent. Currently, we only support a limited number of categories and origins, but if we send all of them, the number of input tokens increase significantly.

In this PR, I update the way we get the origin and category by requesting Gemini to provide the category and origin as a string, in English. We then try to match the value with the taxonomy.

For category, I added a very basic pluralization/singularization system to match to plural/singular form of the noun when the match failed (as the canonical ID can be in plural or singular).

Benefits from this approach:

- We support virtually all categories/origins, as long as we manage to match the value to the taxonomy
- smaller number of input tokens sent to Gemini

Drawbacks:

- The match can fail, in which case we don't have any value
- We potentially lose information when asking to translate to English. For example, in my tests, "Tomates grappes" were translated to "Tomatoes" and "Poires conférence" to "Pears", which is not the most precise category.

## Other possible approach

We ask for Gemini for the category in the original language (French, Spanish,...) and the perform the matching using Open Food Facts taxonomy. The issue is that the category taxonomy currently does not include the full "extended synonyms", which is all synonyms (including plural and singular forms) that are generated and used by Product Opener.
Without this, we can't match "fr:citron-vert" to "en:limes" (while "fr:citrons-verts" can be matched automatically, as it's in the taxonomy `names` for `fr`).

With access to extended synonyms (https://github.com/openfoodfacts/openfoodfacts-server/issues/10742), this would be possible.


open_prices/proofs/ml.py
```diff
@@ -145,109 +147,6 @@ class Products(enum.Enum):
     ZUCCHINI = "en:zucchini"
 
 
-class RawCategory(enum.StrEnum):
```
raphodn: mentioned during this week's call: simplify the PR and only remove the origin enum

raphodn: up :grin:

teolemon:
@copilot resolve the merge conflicts in this pull request

PR #900, opened 2025-08-07
chore: add a script to detect duplicated prices

raphael0202:
Remove duplicated prices, only for prices associated with proof of type
PRICE_TAG.

This command currently only removes duplicated prices coming from the
same proof, duplicated proofs are not handled yet.

We first create a CTE (temporary Common Table Expression) to find all
duplicate prices. Prices are considered as duplicated if their have the
same (`product_code`/`category_tag`, `price`, `proof_id`).

We then loop on each group of duplicate prices, keeping the first one
(the one with the lowest ID) and removing the others. If any of the
duplicate prices is linked to a price tag, we update the price tag to
point to the first price instead of the removed one.

### DIFF
--- a/open_prices/prices/management/commands/remove_duplicates.py
+++ b/open_prices/prices/management/commands/remove_duplicates.py
@@ -1,0 +1,185 @@
+import argparse
+from itertools import groupby
+
+from django.core.management.base import BaseCommand
+
+from open_prices.prices import constants
+from open_prices.prices.models import Price
+from open_prices.proofs.models import PriceTag
+
+
+class Command(BaseCommand):
+    """Remove duplicated prices, only for prices associated with proof of type
+    PRICE_TAG.
+
+    This command currently only removes duplicated prices coming from the
+    same proof, duplicated proofs are not handled yet.
+
+    We first create a CTE (temporary Common Table Expression) to find all
+    duplicate prices. Prices are considered as duplicated if their have the
+    same (`product_code`/`category_tag`, `price`, `proof_id`).
+
+    We then loop on each group of duplicate prices, keeping the first one
+    (the one with the lowest ID) and removing the others. If any of the
+    duplicate prices is linked to a price tag, we update the price tag to
+    point to the first price instead of the removed one.
+    """
+
+    help = "Remove duplicated prices, only for prices with proof of type PRICE_TAG."
+
+    def add_arguments(self, parser: argparse.ArgumentParser) -> None:
+        parser.add_argument(
+            "--apply",
+            action="store_true",
+            help="Run the cleanup and apply changes, otherwise just show what would be done (dry run).",
+            default=False,
+        )
+
+    def handle(self, *args, **options) -> None:  # type: ignore
+        apply = options["apply"]
+
+        self.stdout.write(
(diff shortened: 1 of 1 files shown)


open_prices/prices/management/commands/remove_duplicates.py
raphodn, line 1: for such a command I think i would add tests

open_prices/prices/management/commands/remove_duplicates.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def remove_duplicates(self, price_type: str, apply: bool = False) -> None:
+        if price_type not in constants.TYPE_LIST:
+            self.stdout.write(
+                f"Price type {price_type} is not valid. Valid types are: {constants.TYPE_LIST}"
+            )
+            return
+
+        self.stdout.write(f"Removing duplicates for price type: {price_type}")
+        self.stdout.write("Number of prices before cleanup: %d" % Price.objects.count())
```
raphodn, line 58: add a filter on the price_type ?

open_prices/prices/management/commands/remove_duplicates.py
```diff
@@ -0,0 +1,185 @@
...
+
+        # Create a CTE to find duplicate prices. We use a raw SQL query
+        # to find duplicates based on product_code (or category_tag), price,
+        # and proof_id.
+
+        # we store the product identifier either in `product_code` or
+        # `category_tag` depending on the price type.
+        comparison_field = (
```
raphodn, line 66: if we add a third type this will fail. maybe be more explicit with TYPE_CATEGORY case ?

open_prices/prices/management/commands/remove_duplicates.py
```diff
@@ -0,0 +1,185 @@
...
+            "product_code" if price_type == constants.TYPE_PRODUCT else "category_tag"
+        )
+        duplicate_prices = Price.objects.raw(
```
raphodn, line 69: i would have preferred a django queryset instead ^^

open_prices/prices/management/commands/remove_duplicates.py
raphael0202: Indeed, that would be interesting! How easy is it to test scripts on Django?

open_prices/prices/management/commands/remove_duplicates.py
raphael0202: I now raise an error is the type is not PRODUCT or CATEGORY to prevent this, and make the comment more explicit.

open_prices/prices/management/commands/remove_duplicates.py
raphael0202: Me as well, but it's currently out of my abilities to write such complex Django queries!

open_prices/prices/management/commands/remove_duplicates.py
raphodn: it's doable by calling `from django.core.management import call_command`
alternatively, move the main function `remove_duplicates` to a dedicated file (utils, or better the `moderation` app ?), and test only that function by importing it (create data, call the function, test the result)

raphodn: i can have a look to split the logic in bits if you want

PR #902, opened 2025-08-07
chore: add script to remove duplicate proofs

raphael0202:
Remove duplicated proofs, based on the provided filters.

This script removes all prices and proofs (excluding the reference proof and associated prices) that have the same owner, date, and location ID as the reference proof. It is useful to clean up the database from duplicate proofs that may have been created by mistake.

### DIFF
--- a/open_prices/proofs/management/commands/remove_duplicate_proofs.py
+++ b/open_prices/proofs/management/commands/remove_duplicate_proofs.py
@@ -1,0 +1,88 @@
+import argparse
+
+from django.core.management.base import BaseCommand
+
+from open_prices.prices.models import Price
+from open_prices.proofs.models import Proof
+
+
+class Command(BaseCommand):
+    """Remove duplicated proofs, based on the provided filters.
+
+    This script removes all prices and proofs (excluding the reference proof
+    and associated prices) that have the same owner, date, and location ID as
+    the reference proof. It is useful to clean up the database from duplicate
+    proofs that may have been created by mistake.
+    """
+
+    help = "Remove duplicated proofs, based on a reference proof."
+
+    def add_arguments(self, parser: argparse.ArgumentParser) -> None:
+        parser.add_argument(
+            "--apply",
+            action="store_true",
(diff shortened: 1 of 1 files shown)


open_prices/proofs/management/commands/remove_duplicate_proofs.py
```diff
@@ -0,0 +1,88 @@
...
+    help = "Remove duplicated proofs, based on a reference proof."
+
+    def add_arguments(self, parser: argparse.ArgumentParser) -> None:
+        parser.add_argument(
+            "--apply",
+            action="store_true",
+            help="Run the cleanup and apply changes, otherwise just show what would be done (dry run).",
+            default=False,
+        )
+        parser.add_argument(
```
raphodn, line 27: How did you get the proof_id reference ?

raphodn: A proof with the same owner, date & location is common when users take multiple pictures in the same shop. I think I'm missing something..

raphael0202: Well, this script lacks a bit of context indeed!
It is supposed to solve the mobile app bug that sent the same proof over and over again (up to 730 times), with the same prices. It was only on RECEIPT proofs.

Before running the script:
- checked that no other proof was uploaded at that date in this store by this user
- fetched the ID of the first proof uploaded

You're right that I should at least add filter proofs based on the same type (for the above example, RECEIPT) to prevent unwanted deletions.
But this script is definitely not to be run "as it", it was more of a helper for me to avoid doing the deletion in Django shell (which is more risky in my opinion).

raphael0202: I added the proof type filter mentioned above, and improved the docstring.

raphodn: thanks for the detailed response ! i imagined it was indeed a one-shot, reassured now :)

PR #908, opened 2025-08-08
fix: override `ready_for_validation_price_tag` to False for Smoothie

raphael0202:
On Smoothie v4.20, when uploading a proof,
`ready_for_validation_price_tag` is set to True for all price tags, which was supposed to be done only for bulk proof upload.

See https://github.com/openfoodfacts/smooth-app/issues/6784 for more information about the issue.

Here, we override `ready_for_validation_price_tag` and set it to False when proof upload come from Smoothie v4.20.

Note that it also applied for proof uploaded using Smoothie bulk proof upload, which is an unwanted side effect. Users should upgrade to v4.21 (currently internal) to have `ready_for_validation_price_tag=True` during bulk proof upload.

### DIFF
--- a/open_prices/api/proofs/views.py
+++ b/open_prices/api/proofs/views.py
@@ -1,3 +1,5 @@
+import re
+
 import PIL.Image
 from django.conf import settings
 from django_filters.rest_framework import DjangoFilterBackend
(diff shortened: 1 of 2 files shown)


open_prices/api/proofs/views.py
```diff
@@ -123,8 +145,20 @@ def upload(self, request: Request) -> Response:
             else settings.ANONYMOUS_USER_ID
         )
         source = get_source_from_request(self.request)
+        save_kwargs = {
```
raphodn, line 148: i think i would have done this in the Proof model's `clean` method instead

raphodn: ok for you if i make the change ?

raphael0202: Yes, if you prefer! As this solves one of the biggest duplicate issue, I will merge it now (so that I can deploy it quickly), and let you do the change in a new PR.

PR #919, opened 2025-08-11
chore: use `europe-west1` location

raphael0202:
Solves #917 (finally!)


teolemon:
well done @raphael0202 good catch.

teolemon:
is there a way to use some kind of a wildcard on europe-west* ?

Issue #514, opened 2024-10-11
Detect (and ignore) duplicate proofs


raphodn:
### What

Similar to #422

We would need to store the image hash, and check if a Proof with the same user already exists

### Related issues

- https://github.com/openfoodfacts/open-prices-frontend/issues/534
- https://github.com/openfoodfacts/open-prices/discussions/319
- #422

monsieurtanuki:
I guess same picture dimensions same day same full hash same thumbnail hash same user same location same currency would be enough.

odin-h:
Unless you explicitly include the filename in the data you're hashing, the checksum is only calculated from the file's binary content. Filenames are typically managed by the file system, they are usually not part of the file itself.  Files with the same pixel structure and metadata will generally result in the same output when converted as long as the compression settings are the same (there are exceptions like deflate for png which is default in some versions of the `convert` command from the ImageMagic library, and with variations in implementations between browsers).
But the proofs in the second example you showed aren't actually exact duplicates of each other, they're cropped slightly differently. Whoever uploaded those proofs probably used another app to crop the image each time before importing it again on the proof upload page. Why are the files jpeg though? My proofs are usually webp.

I think relying too much on the hashes of compressed files ties us to only using the current image format and encoding settings in order to detect duplicates. I think a better approach would be to use a perceptual hashing algorithm instead of algorithms like sha or md5 which are mostly used for verifying integrity of files, It will catch some duplicates, but if the image gets cropped or is stripped of its metadata, it will still allow for duplicate uploads. The only problem with perceptual hashes (or "visual fingerprinting") is that they can cause false positives and reject an image you took of a store shelf last week because it looks the same as the ones you took today if it was taken at almost the same angle. Also, the "fuzzy matching" is generally difficult to tune for really close up images of price tags or receipts since they all just look like white backgrounds with text on them to the algorithm.
For those it would probably be better to look at the file creation date (but last I check we didn't save that info), or calculate the hash of the original file in the browser before it's compressed and send that to the server along with the compressed file. But then we can only check for duplicates against what's uploaded after that change is made.

raphael0202:
For information, we use perceptual hashing on Robotoff, and I was really satisfied with the results:
https://github.com/openfoodfacts/robotoff/pull/1272

odin-h:
[...]

PR #920, opened 2025-08-12
fix: check raw image MD5 in script to delete duplicate proofs

raphael0202:
Related issue: #514

This option is activated by default, making the script much safer.

### DIFF
--- a/open_prices/proofs/management/commands/remove_duplicate_proofs.py
+++ b/open_prices/proofs/management/commands/remove_duplicate_proofs.py
@@ -1,9 +1,19 @@
 import argparse
+import hashlib
 
 from django.core.management.base import BaseCommand
 
 from open_prices.prices.models import Price
 from open_prices.proofs.models import Proof
+
+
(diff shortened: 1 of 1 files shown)


teolemon (commented):
Should such a function be in utils ?
Should we run it on new image upload and refuse exact duplicates ?

raphael0202:
> Should such a function be in utils ?

Yes that would be useful to compute a check and add it to DB! When we do this, we can move it to utils.

open_prices/proofs/management/commands/remove_duplicate_proofs.py
```diff
@@ -1,11 +1,21 @@
 import argparse
+import hashlib
 
 from django.core.management.base import BaseCommand
 
 from open_prices.prices.models import Price
 from open_prices.proofs.models import Proof
 
 
+def compute_image_md5(file_path: str) -> str:
```
raphodn, line 10: next step is to store this in the DB so that we can do matching must more easily !

PR #923, opened 2025-08-12
fix: add new environment variables

raphael0202:
- Q2_SYNC: control whether we want to run Q2 tasks synchronously (locally, not for production). Previously, it was set to True if DEBUG was True. However, when launching ML tasks, all of them failed due to timeout. This envvar helps us control more precisely whether we want to keep async tasks even when DEBUG=True.
- GEMINI_API_KEY: For local development, it's easier to use Gemini API instead of Vertex AI (which requires credential JSON files). This envvar allow to use Gemini API if GOOGLE_GENAI_USE_VERTEXAI=false


raphodn (approved):

TTalex:
@raphael0202 Q2_SYNC should probably be set to True in the default file (for local)

For some reason, setting it to False broke all the async_tasks on my machine, not a fun debug 😅

TTalex:
Nope, nothing runs. I can see a print right before the async task, but nothing after.

TTalex:
> The scheduler is running, right?

Today I learned there's a scheduler to start when you're not using docker. Thanks.

raphael0202:
@TTalex the scheduler is run using this command:

```bash
python manage.py qcluster
```

https://github.com/openfoodfacts/open-prices/blob/main/docker-compose.yml#L53

PR #952, opened 2025-08-24
feat(Ml,ReceiptItems): lookup product code using product names

TTalex:
### What

1. Created prices contain unused information about the pair `product_name` - `product_code`
2. When extracting receipt items from the proof prediction, we get product names

- This PR uses both to match detected product names with existing prices and their associated product code
   - Matching rules: exact match on `product_name`, and `location` has to be the same
- The information is stored under `ReceiptItem.predicted_data.predicted_product_code`
- The idea is to use such information to suggest product codes in the Receipt Assistant frontend (example below, work in progress)

image

- I'm not sure about the efficiency of the query, since it requires a new lookup to the price db. But from local perf testing, it's insignificant compared to the AI extraction.

### DIFF
--- a/open_prices/proofs/ml.py
+++ b/open_prices/proofs/ml.py
@@ -28,6 +28,7 @@ from pydantic import BaseModel, Field, computed_field
 
 from open_prices.common import google as common_google
 from open_prices.common import openfoodfacts as common_openfoodfacts
+from open_prices.prices import constants as price_constants
 from open_prices.prices.models import Price
 from open_prices.proofs import constants as proof_constants
 from open_prices.proofs.models import (
@@ -949,7 +950,7 @@ def create_receipt_items_from_proof_prediction(
 ) -> list[ReceiptItem]:
     """Create receipt items from a proof prediction containing receipt item
     detections.
-    Also looks up Prices with similar product name and location to try to
+    Also looks up Prices table for similar product name and location to try to
     extract the product code.
 
     :param proof: the Proof instance to associate the ReceiptItems with
@@ -973,12 +974,20 @@ def create_receipt_items_from_proof_prediction(
     matching_prices = Price.objects.filter(
         product_name__in=product_names,
         location=proof.location,
+        type=price_constants.TYPE_PRODUCT,
     )
     # Using the prices data, create a lookup table matching product names
     # and product codes
-    product_lookup = {
-        price.product_name: price.product_code for price in matching_prices
-    }
+    product_lookup = {}
+    for price in matching_prices:
+        if (
+            price.product_name in product_lookup
+            and product_lookup[price.product_name] != price.product_code
+        ):
+            logger.warning(
+                "Multiple products with the same name found: %s", price.product_name
+            )
+        product_lookup[price.product_name] = price.product_code
 
     created = []
     for index, predicted_item in enumerate(proof_prediction.data.get("items", [])):


TTalex:
> 👌 Niiiiiiice :-) It would be good to have the complete card, beyond barcode (with the product name and the expandable miniature, to get "social proof" that the match is the right one. It could be inline, on hover or on click

Good idea, let's talk about it in the frontend PR that I have not yet created 😄. I'll @ you when it's there.
I'm just waiting to finalize the backend before pushing something, if big changes are required.

open_prices/proofs/ml.py
```diff
@@ -947,17 +948,46 @@ def create_receipt_items_from_proof_prediction(
...
     if proof_prediction.model_name != common_google.GEMINI_MODEL_NAME:
         logger.error(
             "Proof prediction model %s is not a receipt extraction",
             proof_prediction.model_name,
         )
         return []
 
+    # For every predicted item product name, look up a corresponding Price
+    product_names = [
+        item.get("product_name")
```
raphael0202, line 970: Note: I don't know if Gemini sometimes leaves trailing whitespaces, it would be interesting to check (otherwise the match will fail)

open_prices/proofs/ml.py
```diff
@@ -947,17 +948,46 @@ def create_receipt_items_from_proof_prediction(
...
+        for item in proof_prediction.data.get("items", [])
+        if item.get("product_name")
+    ]
+    matching_prices = Price.objects.filter(
+        product_name__in=product_names,
+        location=proof.location,
```
raphael0202, line 976: We should filter to only keep prices of type PRODUCT here, no?

open_prices/proofs/ml.py
```diff
@@ -947,17 +948,46 @@ def create_receipt_items_from_proof_prediction(
...
+    )
+    # Using the prices data, create a lookup table matching product names
+    # and product codes
+    product_lookup = {
```
raphael0202: I would check here that we don't get 2 two different prices, and add a logger.warning (that will be logged by Sentry) in case we have.

raphael0202 (approved):
A few minor remarks, but otherwise good to me :)

open_prices/proofs/ml.py
TTalex: Good idea

open_prices/proofs/ml.py
TTalex: I would expect lots of different prices, and sometimes different products. But that's probably smart to keep track of how often this happens.
That requires rewriting my cool pythonic one-liner, but eh... 😆

open_prices/proofs/ml.py
TTalex: Would the whitespaces be trimmed in DB ? Or saved as is ? Because if gemini does the same mistake of leaving trailing whitespaces every time, it should be fine ? (maybe that's a bad assumption to trust that it would do the same every time)

TTalex:
Note: when price will be created using the suggestion. We should keep track that the source was suggested (aka untrusted).
Also: indexes ?

open_prices/proofs/ml.py
raphael0202: > Would the whitespaces be trimmed in DB ? Or saved as is ? Because if gemini does the same mistake of leaving trailing whitespaces every time, it should be fine ? (maybe that's a bad assumption to trust that it would do the same every time)

The user can also add some whitespace. Currently, we have 2 prices with `product_name` with trailing whitespaces, which is fortunately not a lot:

- `JOUBLETTE BOEUF ECO + `
- `OEUFS  GROS PLEIN `

We should probably trim them on the backend when prices are saved!

PR #964, opened 2025-08-26
feat: prevent duplicate proof upload

### LINKED ISSUE #963: Prevent duplicate proof upload
- Create `md5_hash` field in the `proof` table: this field is used to detect exact duplicate proofs
- Return HTTP 200 with existing proof if the proof is a duplicate (same owner, same location, same type, same date, same MD5 hash)

raphael0202:
Closes #963.


open_prices/proofs/models.py
```diff
@@ -192,13 +192,18 @@ class Proof(models.Model):
     created = models.DateTimeField(default=timezone.now)
     updated = models.DateTimeField(auto_now=True)
 
+    md5_hash = models.CharField(max_length=32, blank=True, null=True)
```
raphodn: to match with the naming of the other fields
```suggestion
    image_md5_hash = models.CharField(max_length=32, blank=True, null=True)
```
and move it up ? (just above `image_thumb_path`)

open_prices/proofs/models.py
```diff
@@ -192,13 +192,18 @@ class Proof(models.Model):
...
+
     objects = models.Manager.from_queryset(ProofQuerySet)()
 
     class Meta:
         # managed = False
         db_table = "proofs"
         verbose_name = "Proof"
         verbose_name_plural = "Proofs"
+        indexes = [
```
raphodn: you can simply add `db_index=True` in the field definition (probably better to delete and rerun the migration)
https://docs.djangoproject.com/en/5.2/ref/models/fields/#db-index

open_prices/proofs/migrations/0022_proof_md5_hash_proof_proofs_md5_has_ee806d_idx.py
raphodn, line 1: I would have separated into 2 PRs
- the creation of the new field (with an extra migration to populate the field for existing proofs)
- and the changes in the api

open_prices/api/proofs/views.py
```diff
@@ -144,10 +146,34 @@ def upload(self, request: Request) -> Response:
...
+            return Response(
+                ProofFullSerializer(duplicate_proof).data,
+                status=status.HTTP_200_OK,
+            )
+
         source = get_source_from_request(self.request)
         save_kwargs = {
             "source": source,
             "owner": owner,
+            "md5_hash": md5_hash,
```
raphodn: it would be better to calculate the md5_hash in the model's save() method.
if proofs are added outside of the API, the field won't be filled

raphodn: the save would then check duplicates, and return a ValidationError, that could be catched by the view

raphodn: I refactored the location API duplicate logic here as an example https://github.com/openfoodfacts/open-prices/pull/965

open_prices/proofs/models.py
raphodn: (ideally one day rename also file_path to image_file_path & same fore mimetype..)

raphodn (approved):

open_prices/api/proofs/views.py
raphodn: (I can refactor later on)

open_prices/proofs/models.py
raphodn: Really sorry because I reread the documentation and they don't recommend anymore 😅 well we'll see when it gets deprecated i guess

raphael0202:
I implemented the switch for image_md5_hash (+updated the index and migration file). Thank you for the review!

Issue #626, opened 2024-12-16
Successful upload of proof image with invalid bearer token


monsieurtanuki:
### What
In off-dart's tests, on the TEST env (`.net`), we try to upload a proof with an invalid bearer token (actual value: `'invalid bearer token'`). It's supposed to fail, but it doesn't anymore.

### Steps to reproduce the behavior:
1. Try to upload a proof with an invalid bearer token in TEST env
2. It works.
3. It shouldn't.

raphodn:
Hi, since last week it's now possible to upload proofs anonymously: https://github.com/openfoodfacts/open-prices/issues/595

monsieurtanuki:
Hi @raphodn!
IMHO "anonymously" and "with an incorrect bearer token" isn't exactly the same.

Anyway, it means that from now on in `uploadProof` for off-dart:
* the bearer token parameter is optional
* if the bearer token is invalid, the call will be successful (today it fails) and will consider that the owner is `openfoodfacts-contributors`

Is that correct?

areebahmeddd:
Is this issue still valid? Should I work on it?

CC: @raphodn

raphodn:
Sure go ahead ! Though it might be a bit complicated for a first dive in the backend code, as it involves DRF stuff 😅

PR #969, opened 2025-08-27
fix(API): Proof upload: if the token is wrong, return a 400 instead of a 201 (and creating it anonymously)

raphodn:
### What

In #607 we started allowing to upload proofs anonymously.

But @monsieurtanuki raised an issue in #626 that users uploading proofs with a wrong token saw their proofs uploaded, but anonymously (the owner was set to `openfoodfacts-contributors`).

This PR fixes this case, by sending back a HTTP 400 if a token is passed (via cookie or header) but it is wrong and unable to authenticate the user.

### DIFF
--- a/open_prices/common/authentication.py
+++ b/open_prices/common/authentication.py
@@ -18,26 +18,32 @@ def get_token_from_cookie(request: Request) -> str | None:
 
 
 def get_token_from_header(request: Request) -> str | None:
-    authorization = request.META.get("HTTP_AUTHORIZATION")  # "Bearer <token>"
-    if authorization and "__U" in authorization:
+    """
+    Expected format: "Bearer <user_id>__U<uuid>"
+    """
+    authorization = request.META.get("HTTP_AUTHORIZATION")
+    if authorization and "Bearer " in authorization:
         return authorization.split(" ")[1]
     return None
 
 
 def has_token_from_cookie_or_header(request: Request) -> bool:
-    return get_token_from_cookie(request) or get_token_from_header(request)
+    return (
+        get_token_from_cookie(request) is not None
+        or get_token_from_header(request) is not None
+    )
 
 
 def get_request_session(request: Request):
     try:
         # If a session cookie is present, use that instead of the header
-        session_cookie = get_token_from_cookie(request)
-        if session_cookie:
-            return get_session(token=session_cookie)
+        token_from_cookie = get_token_from_cookie(request)
+        if token_from_cookie:
+            return get_session(token=token_from_cookie)
 
-        session_header = get_token_from_header(request)
-        if session_header:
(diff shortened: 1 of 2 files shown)


open_prices/api/proofs/views.py
```diff
@@ -140,12 +143,20 @@ def upload(self, request: Request) -> Response:
...
+        # get owner (we allow anonymous upload, only if token is not present)
+        if self.request.user.is_authenticated:
+            owner = self.request.user.user_id
+        else:
+            if has_token_from_cookie_or_header(self.request):
+                return Response(
+                    {
+                        "detail": "Authentication failed. Please pass a valid token, or remove it to upload the proof anonymously."
+                    },
+                    status=status.HTTP_400_BAD_REQUEST,
```
raphodn, line 156: I opted for 400 instead of 403, because the endpoint doesn't *always* require authentication ^^

open_prices/common/authentication.py
```diff
@@ -13,22 +13,32 @@ def create_token(user_id: str) -> str:
...
+    if authorization and "__U" in authorization:
+        return authorization.split(" ")[1]
+    return None
+
+
+def has_token_from_cookie_or_header(request: Request) -> bool:
+    return get_token_from_cookie(request) or get_token_from_header(request)
 
+
+def get_request_session(request: Request):
```
raphodn, line 37: I moved some of the logic to specific methods above

open_prices/common/authentication.py
```diff
@@ -13,22 +13,32 @@ def create_token(user_id: str) -> str:
...
-def get_request_session(request: Request):
+def get_token_from_header(request: Request) -> str | None:
     authorization = request.META.get("HTTP_AUTHORIZATION")  # "Bearer <token>"
-    session_cookie = request.COOKIES.get(settings.SESSION_COOKIE_NAME)
+    if authorization and "__U" in authorization:
+        return authorization.split(" ")[1]
+    return None
+
+
+def has_token_from_cookie_or_header(request: Request) -> bool:
```
raphael0202, line 30: This function returns a string or None, not a bool.
It still works, but for type consistency you should enclose everything under a `bool(...)`

```
>>> ("" or "test") == "test"
True
```

raphael0202 (approved):
A small request change, but otherwise looks good!

open_prices/common/authentication.py
raphodn: Good catch

raphodn: I fixed it and added a bunch of tests in https://github.com/openfoodfacts/open-prices/pull/969/commits/ebfa1f35fea714f4b114cf128bd760e95a8a941d

PR #970, opened 2025-09-03
chore: add anonymize_receipts.py script

raphael0202:


scripts/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+This directory contains scripts related to the Open Prices project.
+
+## `anonymize_receipts.py`
+
+This script is used to anonymize receipt images by redacting personal information such as last names (client or cashier), fidelity card numbers and the hour of the purchase. It utilizes Gemini model to detect personal information, then uses OCR (with Google Cloud Vision) to locate the information in the image. We then redact the identified information from the image and save the anonymized version.
+
+### Installation
+
+Using uv is the recommended way to run the script. First,  [install uv](https://docs.astral.sh/uv/getting-started/installation/).
```
raphodn, line 11: next step : switch the whole project to uv ? :grin:

raphael0202: That would be a good idea! :)
We already use it on Nutripatrol.
uv is particularly nice for scripts imo, as you can set dependencies directly in the file (no need for a requirements.txt)

PR #971, opened 2025-09-03
fix(API): Make the Bearer token prefix case-insensitive

raphodn:
### What

In #969 we did some refactoring on how we read the Bearer token, and involuntarily introduced a case-sensitive check on the word "Bearer".

But the specs allow for flexibility : https://auth0.com/blog/the-bearer-token-case/

This PR fixes this.

Follows a discussion in https://github.com/openfoodfacts/openfoodfacts-dart/issues/1128

### DIFF
--- a/open_prices/common/tests.py
+++ b/open_prices/common/tests.py
@@ -77,6 +77,12 @@ class AuthenticationTest(TestCase):
             headers={"Authorization": f"Bearer {self.user_session.token}"},
         )
         self.assertEqual(get_token_from_header(request), self.user_session.token)
+        # token is case-insensitive
+        request = APIRequestFactory().get(
+            self.url,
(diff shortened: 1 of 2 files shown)


open_prices/common/tests.py
```diff
@@ -77,6 +77,12 @@ def test_get_token_from_header(self):
             headers={"Authorization": f"Bearer {self.user_session.token}"},
         )
         self.assertEqual(get_token_from_header(request), self.user_session.token)
+        # token is case-insensitive
+        request = APIRequestFactory().get(
+            self.url,
+            headers={"Authorization": f"beAreR {self.user_session.token}"},
```
raphael0202, line 83: :smile:

raphael0202 (approved):

PR #973, opened 2025-09-03
fix(Prices): allow to create prices with 3 decimals

raphodn:
### What

Following a discussion in https://github.com/openfoodfacts/smooth-app/issues/6854


open_prices/prices/models.py
```diff
@@ -183,15 +183,15 @@ class Price(models.Model):
 
     price = models.DecimalField(
         max_digits=10,
-        decimal_places=2,
+        decimal_places=3,
```
raphodn, line 186: @raphael0202 should we restrict to 2 for most of the currencies ?

raphodn: todo in a following PR (not priority)

raphodn: so fyi there was an issue in the migration in prod

one price had `price_without_discount=10000100.00` : 10 digits, 2 decimal places. so already the max. so impossible to migrate as this PR adds a decimal place but without adding a new digit... 

I solved it by deleting the price (it was a wrong price) and re-running the migration

raphodn: we'll need to increase the max_digits field.
some countries have currencies with "big amounts", for instance Laos
https://prices.openfoodfacts.org/proofs/21441

image

PR #976, opened 2025-09-07
feat(Prices): track changes in a history table

raphodn:
### What

TODO
- [x] install [django-simple-history](https://django-simple-history.readthedocs.io/en/latest/quick_start.html)
- [x] setup for the Price model (model, migration)
- [x] custom getter & setter (to store the username instead of user id)
- [x] add tests
- [x] apply some changes on the Price model before ? JSONField --> ArrayField : #998 
- [x] init the table `python manage.py populate_history --auto`

### Choices

- store the username (`user_id`) as the history_user
- allow the history_user to be null (when changes coming from a script)

### Todo (other PRs)

- [x] add change reason messages for context : #1016
- API to fetch a price's history ?

### Screenshots

Admin integration

image


raphodn:
> @raphodn This is ready to be reviewed?

Yes :)

PR #978, opened 2025-09-08
fix: return HTTP 201 when uploading duplicate proof on Smooth app

### LINKED ISSUE #977: Return HTTP 201 when smooth app uploads a duplicate proof
The task is considered as a failure if the `/proofs/upload` endpoint returns something else than HTTP 200.
We can return an HTTP 200 status code for all version of Smoothie that are < 4.20.

raphael0202:
Closes #977.

### DIFF
--- a/open_prices/api/proofs/views.py
+++ b/open_prices/api/proofs/views.py
@@ -47,6 +47,12 @@ def is_smoothie_app_version_4_20(source: str | None) -> bool:
     where it sets the `ready_for_price_tag_validation` flag to True when
     uploading price tag proofs, even when it should not be set.
     """
+    return get_smoothie_app_version(source) == (4, 20)
+
+
+def get_smoothie_app_version(source: str | None) -> tuple[int | None, int | None]:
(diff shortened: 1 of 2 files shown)


open_prices/api/proofs/views.py
```diff
@@ -47,17 +47,22 @@ def is_smoothie_app_version_4_20(source: str | None) -> bool:
     where it sets the `ready_for_price_tag_validation` flag to True when
     uploading price tag proofs, even when it should not be set.
     """
+    return get_smoothie_app_version(source) == (4, 20)
+
+
+def get_smoothie_app_version(source: str | None) -> tuple[int | None, int | None]:
```
raphodn, line 53: Does it make sense to move this to openfoodfacts-python later on ?

raphodn (approved):

open_prices/api/proofs/views.py
raphael0202: I don't think so, as the only other backend Smoothie deals with is written in Perl.

PR #980, opened 2025-09-09
chore(Proofs): add a CLI command to compute the MD5 hash of image files

raphael0202:
For proofs for which this field is null.

Linked to #514

Following the addition of the field in #964

### DIFF
--- a/open_prices/proofs/management/commands/compute_missing_image_hash.py
+++ b/open_prices/proofs/management/commands/compute_missing_image_hash.py
@@ -1,0 +1,29 @@
+from pathlib import Path
+
+from django.core.management.base import BaseCommand
+
+from open_prices.proofs.models import Proof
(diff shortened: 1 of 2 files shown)


open_prices/proofs/management/commands/compute_missing_image_hash.py
```diff
@@ -0,0 +1,29 @@
...
+            if not file_path.exists():
+                self.stdout.write(f"File does not exist: {file_path}")
+                continue
+
+            updated += 1
+            proof.image_md5_hash = compute_file_md5_from_path(file_path)
+            self.stdout.write(
+                f"Updated proof {proof.id} with MD5 {proof.image_md5_hash}"
+            )
+            proof.save(update_fields=["image_md5_hash"])
```
raphodn, line 27: :+1: so that we don't update unecessarily the `updated` field

PR #981, opened 2025-09-09
chore: update the remove_duplicate_proofs.py script

raphael0202:
It now uses the new `image_md5_hash` field to find all group of proofs that have the same, date, owner, location_id and image_md5_hash.

### DIFF
--- a/open_prices/proofs/management/commands/remove_duplicate_proofs.py
+++ b/open_prices/proofs/management/commands/remove_duplicate_proofs.py
@@ -1,41 +1,27 @@
 import argparse
-import hashlib
 
 from django.core.management.base import BaseCommand
+from django.db.models import Count
 
 from open_prices.prices.models import Price
 from open_prices.proofs.models import Proof
 
 
-def compute_image_md5(file_path: str) -> str:
-    """Compute the MD5 hash of an image file."""
-    hasher = hashlib.md5()
-    with open(file_path, "rb") as f:
-        while chunk := f.read(8192):
-            hasher.update(chunk)
-    return hasher.hexdigest()
-
-
 class Command(BaseCommand):
-    """Remove duplicated proofs, based on the provided filters.
+    """Remove duplicated proofs.
 
-    This script removes proofs that have the same owner, date,
-    type and location as the provided proof: only the reference proof
-    of the group is kept.
+    This script removes proofs that have the same owner, date, type and
+    location and image MD5 hash: only the first uploaded proof of the group
+    (named reference proof) is kept.
 
     All prices associated with these proofs are also updated to point to the
     reference proof.
 
     It is useful to clean up the database from duplicate proofs that may have
     been created by mistake.
-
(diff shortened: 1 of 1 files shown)


open_prices/proofs/management/commands/remove_duplicate_proofs.py
```diff
@@ -44,65 +30,63 @@ def add_arguments(self, parser: argparse.ArgumentParser) -> None:
...
-        proof_id = options["proof_id"]
-        md5_check = options["md5_check"]
 
-        ref_proof = Proof.objects.get(pk=proof_id)
-        self.stdout.write("=== Running script to remove duplicate proofs...")
-        self.stdout.write(
-            f"Owner: {ref_proof.owner}, Date: {ref_proof.date}, "
-            f"ref proof ID: {ref_proof.id}, Type: {ref_proof.type}, Location ID: "
-            f"{ref_proof.location_id}"
+        # Fetch all proofs for which at least one other proof has the same
```
raphodn, line 37: I'm thinking that we should restrict only on owner & image_md5_hash.
Maybe in a follow-up PR once we discuss this.

For example : Justin regularly batch uploads proofs and mistakenly selects already-uploaded ones. but the location has changed (and maybe the date as well). won't be detected as duplicate. even though it's the exact same image

open_prices/proofs/management/commands/remove_duplicate_proofs.py
```diff
@@ -44,65 +30,63 @@ def add_arguments(self, parser: argparse.ArgumentParser) -> None:
...
+            proofs = list(
+                Proof.objects.filter(
+                    owner=proof_group["owner"],
+                    date=proof_group["date"],
+                    type=proof_group["type"],
+                    location_id=proof_group["location_id"],
+                    image_md5_hash=proof_group["image_md5_hash"],
+                ).order_by("id")
+            )
+            moved_price_count_, deleted_proof_count_ = self.remove_duplicates(
```
raphodn: what's with the underscores at the end ? :sweat_smile:

raphodn:
```suggestion
            proof_group_moved_price_count, proof_group_deleted_proof_count = self.remove_duplicates(
```

open_prices/proofs/management/commands/remove_duplicate_proofs.py
raphael0202: But which one to keep? It's difficult to know which date and/or location is the right one. I would rather display these duplicates in a moderation section, so that either the user or a moderator fixes the issue.

open_prices/proofs/management/commands/remove_duplicate_proofs.py
raphael0202: Just means it's a temporary variable that's not going to be used after, but I'm fine to make it easier to understand with the change you suggested!

PR #983, opened 2025-09-09
fix: return "status": "duplicate" when a duplicate proof is uploaded

raphael0202:
This was part of the original issue
https://github.com/openfoodfacts/open-prices/issues/514#issuecomment-3184397560 but was not implemented.

### DIFF
--- a/open_prices/api/proofs/views.py
+++ b/open_prices/api/proofs/views.py
@@ -193,7 +193,7 @@ class ProofViewSet(
                 status_code = status.HTTP_201_CREATED
 
             return Response(
-                ProofFullSerializer(duplicate_proof).data,
+                {**ProofFullSerializer(duplicate_proof).data, "status": "duplicate"},
                 status=status_code,
             )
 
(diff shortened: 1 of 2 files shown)


open_prices/api/proofs/views.py
```diff
@@ -193,7 +193,7 @@ def upload(self, request: Request) -> Response:
                 status_code = status.HTTP_201_CREATED
 
             return Response(
-                ProofFullSerializer(duplicate_proof).data,
+                {**ProofFullSerializer(duplicate_proof).data, "status": "duplicate"},
```
raphodn: the `status` keyword might not be the best, we use at as field in other models. or it could be confused with the HTTP status.

raphodn: in other custom `Response()` in our code we often use the keyword `detail` instead :information_source:

raphodn: todo : add the same in the Location-duplicate logic

raphael0202: I don't like `detail` much, but indeed `status` could possibly be confused with the HTTP status. I renamed it into `detail` in the meantime.

PR #984, opened 2025-09-09
feat(Proofs): allow to filter by `image_md5_hash`

raphael0202:
As part of #514.

### DIFF
--- a/open_prices/api/proofs/filters.py
+++ b/open_prices/api/proofs/filters.py
@@ -42,6 +42,7 @@ class ProofFilter(django_filters.FilterSet):
     created__lte = django_filters.DateTimeFilter(
         field_name="created", lookup_expr="lte"
     )
+    image_md5_hash = django_filters.CharFilter(field_name="image_md5_hash")
 
     def filter_kind(self, queryset, name, value):
         if value == constants.KIND_COMMUNITY:
--- a/open_prices/api/proofs/filters.py
+++ b/open_prices/api/proofs/filters.py
@@ -42,7 +42,6 @@ class ProofFilter(django_filters.FilterSet):
     created__lte = django_filters.DateTimeFilter(
         field_name="created", lookup_expr="lte"
     )
-    image_md5_hash = django_filters.CharFilter(field_name="image_md5_hash")
 
     def filter_kind(self, queryset, name, value):
         if value == constants.KIND_COMMUNITY:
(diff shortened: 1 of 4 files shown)


open_prices/api/proofs/views.py
```diff
@@ -230,6 +231,16 @@ def process_with_gemini(self, request: Request) -> Response:
         ]
         return Response({"labels": labels}, status=status.HTTP_200_OK)
 
+    @extend_schema(request=MD5HashSerializer)
+    @action(detail=False, methods=["GET"], url_path="check-hash")
```
raphodn:
```suggestion
    @action(detail=False, methods=["GET"], url_path="check-image-md5-hash")
```
to be extra explicit and understand what's the purpose of the endpoint just by reading the url + matches the rest of the code :)

open_prices/api/proofs/views.py
```diff
@@ -230,6 +231,16 @@ def process_with_gemini(self, request: Request) -> Response:
...
+    def check_image_md5_hash(self, request: Request) -> Response:
```
raphodn: an alternate endpoint could be to add the `image_md5_hash` as a possible filter in the `List` view and simply see if it returns a proof or not ? probably as SQL-consuming as this endpoint, but allows to reuse an existing endpoint + returns the full proof ?

open_prices/api/proofs/views.py
raphael0202: You're right! But we don't care whether we use md5 or another algorithm for hashing. I've renamed the function into `check_image_hash` (and I've updated the Serializer class name and route as well).

open_prices/api/proofs/views.py
raphael0202: Indeed, that could be a better idea! At least we would be able to display which proof as the same hash.

raphael0202: I've just pushed a commit that added a new filter instead.

open_prices/api/proofs/filters.py
```diff
@@ -42,6 +42,7 @@ class ProofFilter(django_filters.FilterSet):
     created__lte = django_filters.DateTimeFilter(
         field_name="created", lookup_expr="lte"
     )
+    image_md5_hash = django_filters.CharFilter(field_name="image_md5_hash")
```
raphodn: je push un commit pour simplifier (en basculant le champ directement dans `fields`)

raphodn: ici : https://github.com/openfoodfacts/open-prices/pull/984/commits/e8a8f6c3a9b3d158a976e0f9cdb5db0b5374864d

PR #1005, opened 2025-09-20
fix(Migration): fix price migration 0012 (ArrayField)

TTalex:
### What

Following #998

Note : probably should merge as is, but this fixed my setup

### Why

Some prices have non-array values stored in `origins_tags` & `labels_tags`

### DIFF
--- a/open_prices/prices/migrations/0012_price_labels_tags_temp_price_origins_tags_temp.py
+++ b/open_prices/prices/migrations/0012_price_labels_tags_temp_price_origins_tags_temp.py
@@ -9,7 +9,7 @@ def migrate_labels_tags_data(apps, schema_editor):
     for price in Price.objects.exclude(labels_tags=None).exclude(
         labels_tags__isnull=True
     ):
-        price.labels_tags_temp = price.labels_tags
+        price.labels_tags_temp = price.labels_tags or []
         price.save(update_fields=["labels_tags_temp"])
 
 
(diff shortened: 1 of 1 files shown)


open_prices/prices/migrations/0012_price_labels_tags_temp_price_origins_tags_temp.py
```diff
@@ -18,11 +18,12 @@ def migrate_origins_tags_data(apps, schema_editor):
     for price in Price.objects.exclude(origins_tags=None).exclude(
         origins_tags__isnull=True
     ):
-        price.origins_tags_temp = price.origins_tags
+        price.origins_tags_temp = price.origins_tags or []
```
raphodn: I think instead I would have written
```
if price.origins_tags:
    price.origins_tags_temp = price.origins_tags
    price.save(update_fields=["origins_tags_temp"])
```

raphodn: because when creating the field I set it to `null=True` and without a default

open_prices/prices/migrations/0012_price_labels_tags_temp_price_origins_tags_temp.py
```diff
@@ -43,10 +44,14 @@ class Migration(migrations.Migration):
             ),
         ),
         migrations.RunPython(
-            migrate_labels_tags_data, reverse_code=migrations.RunPython.noop
+            migrate_labels_tags_data,
+            reverse_code=migrations.RunPython.noop,
+            atomic=True,
```
raphodn, line 51: source : https://stackoverflow.com/a/77005775

PR #1012, opened 2025-09-23
feat(Products): New API endpoints to update/add Product Opener products and images

TTalex:
### What

- Backend part of PR https://github.com/openfoodfacts/open-prices-frontend/pull/1685 allowing Product Opener products addition from Open Prices
- Uses two new settings: OFF_DEFAULT_USER & OFF_DEFAULT_PASSWORD, that should contain username and password of a "default" Open Prices user

- The PR also uses two recently-added filters:
   - `source_isnull` on products, to find products added to the prices' database, but not existing on any Product Opener database : https://github.com/openfoodfacts/open-prices/issues/1030
   - `price_id` on proofs, in order to fetch the proof containing a specific price : https://github.com/openfoodfacts/open-prices/issues/1033

### DIFF
--- a/open_prices/api/products/views.py
+++ b/open_prices/api/products/views.py
@@ -7,6 +7,10 @@ from rest_framework.response import Response
 from open_prices.api.products.filters import ProductFilter
 from open_prices.api.products.serializers import ProductFullSerializer
 from open_prices.api.utils import get_object_or_drf_404
+from open_prices.common.openfoodfacts import (
+    update_off_product,
+    update_off_product_image,
+)
 from open_prices.products.models import Product
 
 
@@ -27,3 +31,31 @@ class ProductViewSet(
         product = get_object_or_drf_404(Product, code=code)
         serializer = self.get_serializer(product)
         return Response(serializer.data)
+
+    @action(
+        detail=False, methods=["PATCH", "POST"], url_path=r"off_update/(?P<code>.+)"
+    )
+    def update_off_product(self, request: Request, code):
+        result = update_off_product(
+            code,
+            flavor=request.data.get("flavor", "off"),
+            update_params=request.data.get("update_params"),
+        )
+        if result:
+            return Response(result, status=200)
+        return Response(status=400)
+
+    @action(
+        detail=False,
+        methods=["PATCH", "POST"],
+        url_path=r"off_update_image/(?P<code>.+)",
+    )
+    def update_off_product_image(self, request: Request, code):
+        result = update_off_product_image(
+            code,
+            flavor=request.data.get("flavor", "off"),
+            image_src=request.data.get("image_src"),
+        )
+        if result:
+            return Response(result, status=200)
+        return Response(status=400)
--- a/open_prices/api/products/views.py
+++ b/open_prices/api/products/views.py
@@ -7,6 +7,7 @@ from rest_framework.response import Response
 from open_prices.api.products.filters import ProductFilter
 from open_prices.api.products.serializers import ProductFullSerializer
 from open_prices.api.utils import get_object_or_drf_404
+from open_prices.common.authentication import CustomAuthentication
 from open_prices.common.openfoodfacts import (
     update_off_product,
     update_off_product_image,
@@ -17,6 +18,7 @@ from open_prices.products.models import Product
 class ProductViewSet(
     mixins.ListModelMixin, mixins.RetrieveModelMixin, viewsets.GenericViewSet
 ):
+    authentication_classes = []  # see get_authenticators
     queryset = Product.objects.all()
     serializer_class = ProductFullSerializer
     filter_backends = [DjangoFilterBackend, filters.OrderingFilter]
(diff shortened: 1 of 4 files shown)


open_prices/api/products/views.py
```diff
@@ -27,3 +31,31 @@ def get_by_code(self, request: Request, code):
         product = get_object_or_drf_404(Product, code=code)
         serializer = self.get_serializer(product)
         return Response(serializer.data)
+
+    @action(
+        detail=False, methods=["PATCH", "POST"], url_path=r"off_update/(?P<code>.+)"
+    )
+    def update_off_product(self, request: Request, code):
```
TTalex: Something tells me there should be a serializer somewhere here

open_prices/api/products/views.py
```diff
@@ -27,3 +31,31 @@ def get_by_code(self, request: Request, code):
...
+    @action(
+        detail=False,
+        methods=["PATCH", "POST"],
+        url_path=r"off_update_image/(?P<code>.+)",
+    )
+    def update_off_product_image(self, request: Request, code):
+        result = update_off_product_image(
+            code,
+            flavor=request.data.get("flavor", "off"),
+            image_src=request.data.get("image_src"),
```
TTalex: For simplicity’s sake, I'm feeding image_src as a b64 string representation of the image (pretty easy to generate in the frontend). This should probably be a binary file instead, outside the json representation.

open_prices/common/openfoodfacts.py
```diff
@@ -108,7 +112,7 @@ def get_product(code: str, flavor: Flavor = Flavor.off) -> JSONType | None:
         country=Country.world,
         flavor=flavor,
         version=APIVersion.v2,
-        environment=Environment.org,
+        environment=Environment[os.getenv("ENVIRONMENT")],
```
TTalex: Could be a different PR, but this is quite useful when testing with products in OFF staging.

open_prices/common/openfoodfacts.py
```diff
@@ -277,3 +281,41 @@ def barcode_is_valid(barcode: str) -> bool:
         and len(barcode) >= 6
         and openfoodfacts.barcode.has_valid_check_digit(barcode)
     )
+
+
+def update_off_product(
```
TTalex: Called update, but can also create :)

open_prices/common/openfoodfacts.py
```diff
@@ -277,3 +281,41 @@ def barcode_is_valid(barcode: str) -> bool:
...
+        version=APIVersion.v2,
+        environment=Environment[os.getenv("ENVIRONMENT")],
+    )
+    return client.product.update({"code": code, **update_params})
+
+
+def update_off_product_image(
+    code: str, flavor: str = "off", image_src: str = None, image_field: str = "front"
+) -> JSONType | None:
+    imgdata = base64.b64decode(image_src.split(",")[1])
```
TTalex: See above comment, if the API requires a file instead of a base64 representation, this conversion isn't required.

open_prices/common/openfoodfacts.py
```diff
@@ -277,3 +281,41 @@ def barcode_is_valid(barcode: str) -> bool:
...
+    image_fp = io.BytesIO(imgdata)
+
+    base_url = URLBuilder.world(Flavor(flavor), Environment[os.getenv("ENVIRONMENT")])
+    url = f"{base_url}/cgi/product_image_upload.pl"
```
TTalex: The openfoodfacts python SDK doesn't support image addition.
This code is borrowed from the robotoff repo.
It should probably be added to the SDK, and refactored in both repos.

open_prices/common/openfoodfacts.py
```diff
@@ -277,3 +281,41 @@ def barcode_is_valid(barcode: str) -> bool:
...
+
+    form_data = {}
+
+    form_data["user_id"] = (None, settings.OFF_DEFAULT_USER)
```
TTalex: What if the setting isn't set ?

TTalex:
Add a comment with the original user: (OpenPrices) username

open_prices/api/products/filters.py
```diff
@@ -28,6 +28,9 @@ class ProductFilter(django_filters.FilterSet):
     price_count__lte = django_filters.NumberFilter(
         field_name="price_count", lookup_expr="lte"
     )
+    source_isnull = django_filters.BooleanFilter(
```
raphodn: I'm going to add a commit to change it to `source__isnull` + add a test

open_prices/api/products/views.py
raphodn: decision: @raphael0202 will make changes to `openfoodfacts-python` to be able to use the image v3 api

raphael0202: It's now available in release 3.1.0: https://github.com/openfoodfacts/openfoodfacts-python/pull/389

```
api = API(..., version="v3")
api.product.upload_image(barcode, image_data_base64=image_data_base64)
```

open_prices/common/openfoodfacts.py
raphodn: moved the openfoodfacts-python SDK upgrade in a dedicated PR : #1029

open_prices/api/products/filters.py
raphodn: moved it to a dedicated PR : #1030

open_prices/api/products/views.py
raphodn: later ^^

open_prices/common/openfoodfacts.py
raphodn: moved it to a seperate commit : https://github.com/openfoodfacts/open-prices/commit/89c68994ea2492355e12102a9f3783f9224fa136

open_prices/common/openfoodfacts.py
raphodn: renamed to `create_or_update_product_in_off` :grin:

PR #1027, opened 2025-10-01
refactor(Stats): avoid count fields being None (default 0)

raphodn:
### What

Some count stats where stored as None.
Which gives strange results when ordering by top values.

### Screenshot

Screenshot from 2025-09-26 19-44-21


raphodn:
```
for field in User.COUNT_FIELDS:
   print("User", field)
   print(User.objects.filter(**{field: None}).count())
   print(User.objects.filter(**{f"{field}__isnull": True}).count())

for field in Proof.COUNT_FIELDS:
   print("Proof", field)
   print(Proof.objects.filter(**{field: None}).count())
   print(Proof.objects.filter(**{f"{field}__isnull": True}).count())

for field in Location.COUNT_FIELDS:
   print("Location", field)
   print(Location.objects.filter(**{field: None}).count())
   print(Location.objects.filter(**{f"{field}__isnull": True}).count())

for field in Product.COUNT_FIELDS:
   print("Product", field)
   print(Product.objects.filter(**{field: None}).count())
   print(Product.objects.filter(**{f"{field}__isnull": True}).count())
```

In prod, there are some occurences for `Product.price_currency_count` & `Product.location_type_osm_country_count`

PR #1041, opened 2025-10-07
refactor(Products): Automatically set the uploaded image as the selected front (english)

raphodn:
### What

Following #1012 (APIs to create product & image in OFF)

Fix the image upload to actually select the image as front in OFF

### DIFF
--- a/open_prices/api/products/views.py
+++ b/open_prices/api/products/views.py
@@ -68,6 +68,7 @@ class ProductViewSet(
             code,
             flavor=request.data.get("flavor", Flavor.off),
             image_data_base64=request.data.get("image_data_base64"),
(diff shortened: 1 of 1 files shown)


raphodn:
> Should we forward this decision to the frontend instead ?

Sure, in a v2 at some moment :)
But i feel like it's going to be the expected-and-long path if we start re-creating the OFF product addition directly in OP.
So in the v1 keep it simple ? with presets like this, we could add a message informing the user that the product will be in English + the image will be the front.

raphodn:
this is temporary fix until we give the user the ability to set the type of image directly in the frontend.
the experiment has only been communicated to a handful of "very active" users.

PR #1045, opened 2025-10-08
fix: normalize barcodes in prices and products tables

### LINKED ISSUE #1044: Normalize barcodes before saving the price
Currently, to my understanding, no barcode normalization is performed before saving the price.

It means that the same product can be saved under two different barcodes.

It is the case for example for this product: [0000059032823](https://prices.openfoodfacts.org/api/v1/products?code=0000059032823) and [59032823](https://prices.openfoodfacts.org/api/v1/products?code=59032823).
The second barcode is the right one (EAN8), and can be found in the Open Food Facts database.
The first item was added in the `products` table after a user added the price for this product.

We should therefore normalize the barcode before saving the price, and run a script to deduplicate the `products` and `prices` table using the `normalize_barcode` function from the Python SDK.

raphael0202:
Closes #1044.

### DIFF
--- a/open_prices/prices/management/commands/normalize_barcodes.py
+++ b/open_prices/prices/management/commands/normalize_barcodes.py
@@ -1,0 +1,69 @@
+import argparse
+
+from django.core.management.base import BaseCommand
+from openfoodfacts import normalize_barcode
+
+from open_prices.prices.models import Price
+from open_prices.products.models import Product
+
+
+class Command(BaseCommand):
+    """Normalize barcodes for the `products` and `prices` tables.
+
+    This command will go through all `product_code` in the `prices` table and
+    `code` in the `products` table, and normalize the barcodes (remove leading
+    zeros, pad to 8 or 13 digits).
+    """
+
+    help = "Normalize barcodes for the `products` and `prices` tables."
+
+    def add_arguments(self, parser: argparse.ArgumentParser) -> None:
+        parser.add_argument(
+            "--apply",
+            action="store_true",
+            help="Run the normalization and apply changes, otherwise just show what would be done (dry run).",
+            default=False,
+        )
+
+    def handle(self, *args, **options) -> None:  # type: ignore
+        apply = options["apply"]
+
+        self.stdout.write("=== Running script to normalize barcodes...")
+        if not apply:
+            self.stdout.write("Running in dry run mode. Use --apply to apply changes.")
+
+        self.normalize_barcodes(apply=apply)
+
+    def normalize_barcodes(self, apply: bool = False) -> None:
+        for price in Price.objects.has_type_product().all():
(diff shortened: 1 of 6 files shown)


open_prices/prices/management/commands/normalize_barcodes.py
```diff
@@ -0,0 +1,69 @@
...
+                        f"Price ID {price.id}: Normalizing barcode from {price.product_code} to {normalized_code}"
+                    )
+                    if apply:
+                        price.product_code = normalized_code
+                        # we also update `product_id` as we changed the
+                        # barcode, and the product ID is updated automatically
+                        # in `save()`
+                        price.save(update_fields=["product_code", "product_id"])
+
+        for product in Product.objects.all():
```
raphodn, line 52: this might take a loooong time on 4M products, no ?

raphodn: an idea is to narrow down the queryset ?
for instance hide Products with a barcode length of 13 that doesn't start with a 0 ?

raphodn:
```
from django.db.models.functions import Length

Product.objects.count()
// 4105391

Product.objects.annotate(code_length_annotated=Length("code")).exclude(Q(code_length_annotated=13) & ~Q(code__startswith="0")).exclude(Q(code_length_annotated=8) & ~Q(code__startswith="0"))
// 1171675
// source breakdown: {'off': 1156929, 'obf': 7594, 'opf': 3815, None: 1689, 'opff': 1648}

```

raphodn: also, if you say that `normalize_barcode` is run in OxF, then we should only run it on OP products, no ? (with source = None)

raphodn: hmm actually i just ran some tests, we have in our DB many products coming from OxF that are not normalized
```
Counter({'off': 66573, 'obf': 532, None: 530, 'opf': 220, 'opff': 40})
```

raphodn: the tests run quite fast even on the ~1M products isolated above, so I'll keep the queryset as-is :+1:

raphael0202: Can you provide a few examples with source=OxF where barcodes are not normalized?
I believe they were imported before we applied barcode normalization on Product Opener.

raphael0202: And this script is intended to be run only once :)

raphodn: > I believe they were imported before we applied barcode normalization on Product Opener.

when was it applied on Product Opener ?

```
00746816400122002829
4206600
4312
...
```

PR #1047, opened 2025-10-09
refactor: following normalize PR, add test on command and additional cleanup

raphodn:
### What

Following #1045 
- add a test on the `normalize_barcodes` command
- move the `normalize_code` action in the save instead of the clean (keep the clean for validation)

### DIFF
--- a/open_prices/prices/tests.py
+++ b/open_prices/prices/tests.py
@@ -328,21 +328,32 @@ class PriceModelSaveTest(TestCase):
             )
 
     def test_product_code_normalization_on_save(self):
-        # barcode of length 12 gets padded to 13
-        price = PriceFactory(product_code="123456789100")
-        self.assertEqual(price.product_code, "0123456789100")
-        # barcode of length 14 with leading zero gets trimmed to 13
-        price = PriceFactory(product_code="00123456789100")
-        self.assertEqual(price.product_code, "0123456789100")
-        # EAN8 without leading 0 stays the same
-        price = PriceFactory(product_code="51234567")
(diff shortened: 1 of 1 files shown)


open_prices/prices/models.py
```diff
@@ -304,11 +315,6 @@ def clean(self, *args, **kwargs):
                     "product_code",
                     "Should not be a boolean or an invalid string",
                 )
-            if self.product_code.isdigit():
```
raphodn, line 307: i moved it outside of clean, in a dedicated method

open_prices/products/models.py
```diff
@@ -124,22 +124,24 @@ def set_default_values(self):
             if getattr(self, field_name) is None:
                 setattr(self, field_name, [])
 
+    def normalize_code(self):
```
raphodn, line 127: same here

open_prices/prices/tests.py
```diff
@@ -327,21 +328,32 @@ def test_price_product_validation(self):
...
-        # EAN8 without leading 0 stays the same
-        price = PriceFactory(product_code="51234567")
-        self.assertEqual(price.product_code, "51234567")
-        # barcode of length < 8 gets padded to 8
-        price = PriceFactory(product_code="4567")
-        self.assertEqual(price.product_code, "00004567")
         # barcode of length > 13 without leading 0 stays the same
         price = PriceFactory(product_code="8658585456785867")
         self.assertEqual(price.product_code, "8658585456785867")
+        # barcode with letters stays the same
```
raphodn, line 352: reorganized the tests + added extra with letters inside

PR #1048, opened 2025-10-09
feat(Proofs): track changes in a history table

raphodn:
### What

Similar to #976 (Price history)


open_prices/proofs/models.py
```diff
@@ -196,6 +197,15 @@ class Proof(models.Model):
     created = models.DateTimeField(default=timezone.now)
     updated = models.DateTimeField(auto_now=True)
 
+    history = HistoricalRecords(
+        excluded_fields=COUNT_FIELDS,
```
raphodn, line 201: difference with Price history: here we exclude the count fields, that can be incremented regularly, but are not interesting for history purpose. As they would create history entries on every change..

raphodn: Note: django-simple-history creates a history entry on each `save()`.
It's overly complex (docs [here](https://django-simple-history.readthedocs.io/en/3.5.0/querying_history.html#save-without-a-historical-record)) to avoid an entry even if there are no diffs (for instance only a count has changed, but because it's excluded, there won't be any diffs...).

IMO it is better to set up a "duplicate cleanup" command in cron instead (in another PR)

raphodn: done here: #1051

PR #1057, opened 2025-10-10
refactor(History): new method to get Price & Proof instance history list

raphodn:
### What

- new `history.build_instance_history_list(instance)` helper method
- used when calling `price.get_history_list()` & `proof.get_history_list()`

### Why

will be useful to format the API history endpoint (see #1058)


raphael0202 (approved):
👍

PR #1068, opened 2025-10-12
fix(Proofs): Prevent exception if there is any problem with the img thumbnail processing

lucasvazq:
### What

Prevents exceptions when processing an Image. \
If an error occurs, the function simply returns `None`. This is acceptable since `None` is an explicitly allowed return type according to the type annotation.

### Screenshot

Not applicable

### Fixes bug(s)

#942 

### Part of 

Standalone issue/PR

### DIFF
--- a/open_prices/proofs/utils.py
+++ b/open_prices/proofs/utils.py
@@ -83,28 +83,28 @@ def generate_thumbnail(
     if not mimetype.startswith("image"):
         return None
     file_full_path = generate_full_path(current_dir, file_stem, extension)
-    try:
-        with Image.open(file_full_path) as img:
+    with Image.open(file_full_path) as img:
+        try:
             img_thumb = img.copy()
-            # set any rotation info
-            img_thumb = ImageOps.exif_transpose(img)
-            # transform into a thumbnail
-            img_thumb.thumbnail(thumbnail_size, Image.Resampling.LANCZOS)
-            image_thumb_full_path = generate_full_path(
-                current_dir, f"{file_stem}.{settings.THUMBNAIL_SIZE[0]}", extension
-            )
-            # avoid 'cannot write mode RGBA as JPEG' error
-            if mimetype in ("image/jpeg",) and img_thumb.mode in ("RGBA", "P"):
-                img_thumb = img_thumb.convert("RGB")
-            # save (exif will be stripped)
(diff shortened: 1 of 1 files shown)


open_prices/proofs/utils.py
```diff
@@ -80,9 +80,10 @@ def generate_thumbnail(
     thumbnail_size: tuple[int, int] = settings.THUMBNAIL_SIZE,
 ) -> str | None:
     """Generate a thumbnail for the image at the given path."""
```
lucasvazq, line 82: Add an early return, then wrap the `.copy` method with a try/except block.

Most of the changes are due to indentation differences.

open_prices/proofs/utils.py
```diff
@@ -102,7 +103,8 @@ def generate_thumbnail(
                 f"{file_stem}.{settings.THUMBNAIL_SIZE[0]}",
                 extension,
             )
-    return image_thumb_path
+    except (OSError, Exception):
```
raphael0202: Catching an `Exception` is too broad here. And I would prefer only enclosing image loading between a try/except, and letting the process crashes (and the error logged in Sentry) for the rest of the function.

raphael0202 (changes requested):
Thank you for the PR! I added a comment

open_prices/proofs/utils.py
lucasvazq: @raphael0202 

I'll give you two options, tell me which one you prefer.

## Op 1

I can simplify the current code by removing the added try/except block and implementing the following change:
```py
with Image.open(file_full_path) as img:
    try:
        img_thumb = img.copy()
    except OSError:
        return None

    # the rest of the code
    ...
```
With this approach, we can catch the error precisely where Sentry reported it in the issue.

## Op 2

Alternatively, we could go for another solution:
```py
with Image.open(file_full_path) as img:
    try:
        img.seek(0)
        img.load()
    except (EOFError, OSError):
        # EOFError: by seek
        # OSError: by load
        return None

    # the rest of the code
    ...
```
This approach selects and decodes the first frame of any image, whether animated (such as a GIF) or static (such as a PNG). It ensures that the frame used is valid and not corrupted.

raphael0202: Option 1 looks better to me!

raphael0202:
Thank you!

PR #1022, opened 2025-09-27
feat(PriceTag): fix some barcodes post-prediction with custom rules

raphodn:
### What

Linked to https://github.com/openfoodfacts/open-prices/issues/1020

PR #1085, opened 2025-10-18
refactor(PriceTag): post-prediction, normalize barcodes

raphodn:
### What

Similar to #1022 (where we added USD-specific post-prediction barcode cleaning)
And following https://github.com/openfoodfacts/open-prices/issues/1044

Also normalize predicted barcodes

### DIFF
--- a/open_prices/proofs/ml.py
+++ b/open_prices/proofs/ml.py
@@ -19,6 +19,7 @@ import typing_extensions as typing
 from asgiref.sync import async_to_sync
 from django.conf import settings
 from google import genai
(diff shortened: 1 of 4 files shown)


open_prices/proofs/ml.py
raphael0202, line 915: We use directly response.parsed to fill the data here. Shoudn't we update the barcode here?

raphodn: Oh yeah..
But then how come the US barcode stuff works ? 😅

PR #1102, opened 2025-10-27
fix: set timeout properly for scheduled task

raphael0202:
Following #1100

Contrary to the example in the documentation
(https://django-q2.readthedocs.io/en/master/schedules.html#schedule), passing timeout through `q_options` is ignored.

This was something that was already present in django-q package (https://github.com/Koed00/django-q/issues/332).

Passing timeout as a keyword argument works (I tested it), this fix is implemented here.

### DIFF
--- a/open_prices/common/tasks.py
+++ b/open_prices/common/tasks.py
@@ -170,6 +170,6 @@ for task_name, (task_cron, q_options) in CRON_SCHEDULES.items():
             name=task_name,
             schedule_type=Schedule.CRON,
             cron=task_cron,
-            q_options=q_options,
+            **q_options,
         )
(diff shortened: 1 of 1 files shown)


open_prices/common/tasks.py
```diff
@@ -170,6 +170,6 @@ def history_cleanup_task():
             name=task_name,
             schedule_type=Schedule.CRON,
             cron=task_cron,
-            q_options=q_options,
+            **q_options,
```
raphodn, line 173: we might be able to move to django-tasks once it lands in Django 6.
https://docs.djangoproject.com/en/dev/topics/tasks/

dunno if it'll have all the features to effectively migrate or not

PR #1108, opened 2025-11-02
fix(Products): convert 'product_quantity' field to Decimal instead of Integer

### LINKED ISSUE #1104: Decimals in source quantity are not used for price-per quantity
### What

For products with a quantity with decimal values (e.g., “8.6 g”), Prices will drop the decimals (not even attempt to round), and will not include them in “price per quantity” calculations, leading those to be wrong.

1. Go to https://prices.openfoodfacts.org/products/8720289261542
2. Note the quantity is shown as “8 g” when the product on OBF is “8.6 g”
3. Note the calculated price-per-quantity is “SEK 1,250.00 / kg” rather than “SEK 1,162.79 / kg”.

### Expected behavior

I expected the price-per-quantity to be based on the actual known quantity rather than an approximation of it.

### Part of
- https://github.com/openfoodfacts/open-prices/issues/77

raphodn:
### What

Closes #1104

cc @Freso


open_prices/products/models.py
```diff
@@ -90,7 +90,12 @@ class Product(models.Model):
 
     product_name = models.CharField(blank=True, null=True)
     image_url = models.CharField(blank=True, null=True)
-    product_quantity = models.IntegerField(blank=True, null=True)
+    product_quantity = models.DecimalField(
+        max_digits=10,
+        decimal_places=3,
```
raphodn, line 95: how many decimal_places? Can we find the exact code in https://github.com/openfoodfacts/openfoodfacts-server where this is set?

raphodn: Ok so there are many edge cases where the number of decimals is much more than just 3.
For instance the oz to g conversion...

raphodn: @monsieurtanuki will there be an impact on the mobile app if the Product.product_quantity field returned by OP is now a decimal ?

monsieurtanuki: @raphodn I guess it will crash the app each time we find a decimal quantity.
If now you return `int` (e.g. `3`) as `double` (e.g. `3.0`) it will crash systematically
```dart
  @JsonKey(name: 'product_quantity')
  int? quantity;
```

monsieurtanuki: @raphodn Beyond that decimal issue, wouldn't it make sense to add a unit to the value?

raphodn: this is data synced with OFF. we have `product_quantity` & `product_quantity_unit`.
we're thinking of syncing `quantity` as well and doing the parsing ourselves, to allow managing non g/L units (like egg quantity, or oz...)

monsieurtanuki: > this is data synced with OFF. we have `product_quantity` & `product_quantity_unit`. we're thinking of syncing `quantity` as well and doing the parsing ourselves, to allow managing non g/L units (like egg quantity, or oz...)

Fair enough.

Fun fact: in Smoothie we don't use `productQuantity` and `productQuantityUnit` (neither from OFF nor Prices), but that will still crash the app. Actually "crash" probably means here "no results found / error".

We definitely need a more resilient json decoding in off-dart. That wouldn't help in the short term, given our slow roll out system and the users' enthusiasm about upgrading the app.

raphodn: > I guess it will crash the app each time

what we can do is that depending on the app_version the backend returns an Integer or a Decimal

monsieurtanuki: > what we can do is that depending on the app_version the backend returns an Integer or a Decimal

That would be ugly and source of confusion. Would make little sense for other apps.
That would probably work, though, but it's not the silver bullet we should rely on in the future.

raphodn: i'd rather have this sort of mecanism, rather then bumping the API version for every change..
it could be stored and documented in a dedicated file.
we already have a couple of rules in that matter with different versions of Smoothie depending on the break.
and X years later clean the code.

monsieurtanuki: Whatever works.

raphodn: I thought about it overnight, would you prefer Open Prices API starting to version the API then ? I think I'm not comfortable with this solution because I'm scared of its complexity. But I can look into it, a good first test with this group of changes. Bump to v1.1 alright ?

https://www.django-rest-framework.org/api-guide/versioning/

monsieurtanuki: > would you prefer Open Prices API starting to version the API then ?

@raphodn Oh, I thought you weren't keen on "bumping the API version for every change", but of course that would be the best solution, from my point of view.
Could indeed be a good starting point to understand the API versioning process, with that minor issue/PR with little extra code.

Freso: FWIW, I think bumping API version whenever there’s a breaking change (similar to how product opener does) is the more sane future-proof approach to take, esp. as the amount of clients/reusers grow. Using our own API versioning in `if` clauses etc. also makes more sense than handling an (hopefully) ever-growing list of clients and their `app_version`s.

raphodn: I created a dedicated discussion to explore this subject: https://github.com/openfoodfacts/open-prices/discussions/1283

PR #1123, opened 2025-11-15
feat(Moderation): new endpoint to list all flags (restricted to moderators)

raphodn:
### What

New GET `/flags` endpoint allowing moderators to list all the flags
- restricted to moderators
- with filtering & ordering

### Todo (futur PRs)

- return full price / proof object ?
- edit a flag status

### Screenshots

|API filters|API response|
|-|-|
|image|image|

### DIFF
--- a/open_prices/api/moderation/filters.py
+++ b/open_prices/api/moderation/filters.py
@@ -1,23 +1,23 @@
 import django_filters
 
-from open_prices.moderation.constants import FLAG_ALLOWED_CONTENT_TYPE_LIST
+from open_prices.moderation import constants as moderation_constants
 from open_prices.moderation.models import Flag, FlagReason
 
 
 class FlagFilter(django_filters.FilterSet):
+    # in the serializer we replace the content_type field with model upper
     content_type = django_filters.MultipleChoiceFilter(
         field_name="content_type__model",
-        lookup_expr="iexact",
-        choices=[
-            (model.upper(), model)
-            for app_label, model in FLAG_ALLOWED_CONTENT_TYPE_LIST
-        ],
+        choices=moderation_constants.FLAG_ALLOWED_CONTENT_TYPE_CHOICES_UPPER,
+        method="filter_content_type",
     )
     reason = django_filters.MultipleChoiceFilter(
-        field_name="reason",
-        choices=FlagReason.choices,
+        field_name="reason", choices=FlagReason.choices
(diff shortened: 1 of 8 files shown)


open_prices/api/moderation/serializers.py
```diff
@@ -13,8 +13,11 @@ class Meta:
...
 class FlagSerializer(serializers.ModelSerializer):
-    content_type_display = serializers.ReadOnlyField()
+    content_type = serializers.SerializerMethodField()
 
     class Meta:
         model = Flag
         fields = "__all__"
+
+    def get_content_type(self, obj):
+        return obj.content_type_display.upper()
```
raphael0202: isn't it `obj.content_type.upper()`?

open_prices/api/moderation/filters.py
```diff
@@ -0,0 +1,23 @@
+import django_filters
+
+from open_prices.moderation.constants import FLAG_ALLOWED_CONTENT_TYPE_LIST
+from open_prices.moderation.models import Flag, FlagReason
+
+
+class FlagFilter(django_filters.FilterSet):
+    content_type = django_filters.MultipleChoiceFilter(
+        field_name="content_type__model",
+        lookup_expr="iexact",
```
raphael0202: Why do we need an `iexact lookup here?

raphael0202 (changes requested):

open_prices/api/moderation/serializers.py
raphodn: fixed in https://github.com/openfoodfacts/open-prices/pull/1123/commits/becb0cc463b7e757cf06fb68bf5ebef4b9dd9dce
- put `obj.content_type_display`
- added a test

open_prices/api/moderation/filters.py
raphodn: the idea is to replace in the API (serializer and here the filter) the default `content_type` (internal id) with a readable `content_type_display` ("PRICE", "PROOF"). I'm going to see if there's an alternative solution

raphodn: refactored stuff here : https://github.com/openfoodfacts/open-prices/pull/1123/commits/a3d8b63d5e7e285111ef9ad86a2b3063791501c7
- switched to an explicit queryset
- also used in the admin
- added tests

PR #1128, opened 2025-11-20
feat: improve price tag detection

raphael0202:
refactor(ml): use new preprocessing pipeline implemented in Python SDK

See https://github.com/openfoodfacts/openfoodfacts-python/pull/400
for more information.

It's much faster than the previous preprocessing pipeline, and is
the same as the one from Ultralytics.

---

 fix: decrease object detection threshold to 0.25

0.5 is too high for object detection, and it's better to detect a bogus
price tag than to miss one.

### DIFF
--- a/open_prices/proofs/ml.py
+++ b/open_prices/proofs/ml.py
@@ -721,7 +721,7 @@ def detect_price_tags(
     label_names: list[str] = PRICE_TAG_DETECTOR_LABEL_NAMES,
     image_size: int = PRICE_TAG_DETECTOR_IMAGE_SIZE,
     triton_uri: str = settings.TRITON_URI,
-    threshold: float = 0.5,
+    threshold: float = 0.25,
 ) -> ObjectDetectionRawResult:
     """Detect the price tags in a proof image.
 
@@ -734,6 +734,7 @@ def detect_price_tags(
     :param image_size: the size of the image, defaults to IMAGE_SIZE
     :param triton_uri: the URI of the Triton server, defaults to
         settings.TRITON_URI
+    :param threshold: the detection threshold, defaults to 0.25
     :return: the detection results
     """
     detector = ObjectDetector(


open_prices/proofs/ml.py
```diff
@@ -720,7 +721,7 @@ def detect_price_tags(
     label_names: list[str] = PRICE_TAG_DETECTOR_LABEL_NAMES,
     image_size: int = PRICE_TAG_DETECTOR_IMAGE_SIZE,
     triton_uri: str = settings.TRITON_URI,
-    threshold: float = 0.5,
+    threshold: float = 0.25,
```
raphodn, line 724: i would have put this in a separate PR

pyproject.toml
```diff
@@ -58,7 +58,7 @@ factory-boy = "^3.3.3"
 google-genai = "~1.46.0"
 gunicorn = "^23.0.0"
 ipython = "^9.6.0"
-openfoodfacts = {version = "~3.1.0", extras = ["ml", "redis"]}
+openfoodfacts = {version = "~3.3.0", extras = ["ml", "redis"]}
```
raphodn, line 61: since the upgrade to 3.3.0, i have issues installing the dependencies locally

```
- Installing perf-analyzer (2.59.1): Failed

| Unable to find installation candidates for perf-analyzer (2.59.1)
| 
| This is likely not a Poetry issue.
| 
|   - 2 candidate(s) were identified for the package
|   - 2 wheel(s) were skipped as your project's environment does not support the identified abi tags
| 
| The following wheel(s) were skipped as the current project environment does not support them due to abi compatibility issues.
| 
|   - perf_analyzer-2.59.1-py3-none-manylinux_2_38_aarch64.whl
|   - perf_analyzer-2.59.1-py3-none-manylinux_2_38_x86_64.whl
| 
| If you would like to see the supported tags in your project environment, you can execute the following command:
| 
|     poetry debug tags
| 
| Solutions:
| Make sure the lockfile is up-to-date. You can try one of the following;
| 
|     1. Regenerate lockfile: poetry lock --no-cache --regenerate
|     2. Update package     : poetry update --no-cache perf-analyzer
| 
| If neither works, please first check to verify that the perf-analyzer has published wheels available from your configured source that are compatible with your environment- ie. operating system, architecture (x86_64, arm64 etc.), python interpreter.
```

raphodn: Maybe linked to https://docs.nvidia.com/deeplearning/triton-inference-server/archives/triton-inference-server-2620/release-notes/rel-25-10.html

> Perf Analyzer is no longer part of the client released archive and can be installed separately using pip install perf-analyzer

raphodn: maybe linked to my Ubuntu version ? I have 22.04

and manylinux_2_38 seems to require at least Ubuntu 23
https://github.com/mayeut/pep600_compliance

raphael0202: Did `poetry debug tags` provide useful info?

raphodn: Yes I saw that I was missing the latest ones. I'll see if I can upgrade to Ubuntu 24 over the holidays..

PR #1130, opened 2025-11-21
refactor(ml): split open_prices.proofs.ml into submodules

raphael0202:
- classification
- ocr
- price_tags
- receipts

This makes the module easier to manage.

### DIFF
--- a/open_prices/proofs/ml/__init__.py
+++ b/open_prices/proofs/ml/__init__.py
@@ -1,0 +1,55 @@
+"""
+Proof ML/AI
+- predict Proof type with triton
+- detect Proof's PriceTags with triton
+- extract data from PriceTags with Gemini
+"""
+
+import logging
+from pathlib import Path
+
+from PIL import Image
+
+from open_prices.proofs.ml.classification import run_and_save_proof_type_prediction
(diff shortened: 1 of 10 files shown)


open_prices/proofs/ml/__init__.py
raphodn, line 1: not a fan of putting code in `__init__`

open_prices/proofs/ml/price_tags.py
raphodn, line 1: an alternative could have been to move the ML logic in a dedicated app (at the same level as proof), to help document it, and in the future have non-proof ML

open_prices/proofs/ml/__init__.py
raphael0202: I agree it makes the code a bit harder to discover, but I didn't come up with a good module name for the high level functions.

raphael0202: I agree it makes the code a bit harder to discover, but I didn't come up with a good module name for the high level functions.

open_prices/proofs/ml/price_tags.py
raphael0202: That's a good idea!
We will definitely do this once we start to do ML for things other than proofs. In the meantime, we can leave it like this :)

PR #1138, opened 2025-11-27
feat(Proofs): create a cropped image of the price tag after DB instance creation

### LINKED ISSUE #1137: Store the price tags individual images on server
Currently, we don't store the image of each cropped price tag on disk: we only store the full image, with the associated bounding box of the price tag.

It has a number of drawbacks:
- on the validation assistant, we have to download the full image, then perform the crop. It results in higher bandwidth and latency, and possibly higher RAM usage (because we perform the crop client-side)
- we cannot easily use the price tag image on other services: we need to use the Robotoff crop API to get it

The latter point is a current blocker for me: as part of the [llm-evals](https://github.com/openfoodfacts/openfoodfacts-ai/tree/develop/llm-evals) package, I need the LLM provider to be able to fetch the price tag images. As Robotoff API is currently blocked by robots.txt, it's currently not possible to use Robotoff for image cropping.

The only drawback of this solution is the higher storage need, but I find it acceptable.

## Proposal

Store the price tag of ID `200000` in `/img/price-tags/000/200/000/000200000.webp`, which would allow us to store up to 1B price tags (still a long way to go ;))

When the price tag is deleted, delete the associated image. When the bounding box is update, regenerate the price tag image.

## Todo

- [x] store the cropped images : #1138
- [x] script to run on the history (and generate the cropped images)
- [x] return the cropped images in the API (serializer only, not a field)

raphael0202:
Closes #1137.

### DIFF
--- a/open_prices/proofs/utils.py
+++ b/open_prices/proofs/utils.py
@@ -296,7 +296,7 @@ def get_price_tag_image_path(price_tag_id: int) -> str:
     digits each. The image extension is .webp.
 
     Example: for price_tag_id = 200000, the path will be:
-    /img/price-tags/000/200/000/000200000.webp
+    /img/price-tags/000/200/000200000.webp
 
     :param price_tag_id: The ID of the price tag.
     :return: The full path to the price tag image.
@@ -304,10 +304,9 @@ def get_price_tag_image_path(price_tag_id: int) -> str:
     id_str = str(price_tag_id).zfill(9)
     part1 = id_str[0:3]
     part2 = id_str[3:6]
-    part3 = id_str[6:9]
 
     filename = f"{id_str}.webp"
-    relative_path = os.path.join("price-tags", part1, part2, part3, filename)
+    relative_path = os.path.join("price-tags", part1, part2, filename)
     return os.path.join(settings.IMAGES_DIR, relative_path)
 
 
@@ -318,7 +317,7 @@ def generate_price_tag_image(price_tag: PriceTag) -> None:
     determine the file path.
 
     The structure of the path is as follows:
-    /img/price-tags/000/200/000/000200000.webp (for price_tag_id = 200000)
+    /img/price-tags/000/200/000200000.webp (for price_tag_id = 200000)
 
     :param price_tag: The price tag object containing the proof and bounding
         box.


open_prices/proofs/utils.py
```diff
@@ -282,3 +286,58 @@ def match_receipt_item_with_price(receipt_item: ReceiptItem, price: Price) -> bo
...
+
+    try:
+        cropped_img = crop_image(proof_image_path, price_tag.bounding_box)
+        output_path = get_price_tag_image_path(price_tag.id)
+
+        # Ensure output directory exists
+        os.makedirs(os.path.dirname(output_path), exist_ok=True)
+
+        # Save as WebP
+        cropped_img.save(output_path, "WEBP")
```
raphodn, line 340: if the original proof image is not webp, will it be converted ?

open_prices/proofs/models.py
```diff
@@ -629,6 +629,34 @@ def get_predicted_product_name(self):
...
+        generate_price_tag_image(instance)
+
+
+@receiver(signals.post_delete, sender=PriceTag)
+def price_tag_post_delete_remove_image(sender, instance, **kwargs):
+    import os
+
+    from open_prices.proofs.utils import get_price_tag_image_path
+
+    image_path = get_price_tag_image_path(instance.id)
```
raphodn, line 652: shouldn't we store this in a field instead ?
so that the frontend (or any API user) can easily retrieve and display the images ?

open_prices/proofs/utils.py
```diff
@@ -282,3 +286,58 @@ def match_receipt_item_with_price(receipt_item: ReceiptItem, price: Price) -> bo
...
+
+
+def generate_price_tag_image(price_tag: PriceTag) -> None:
+    """Crop the price tag from the proof image and save it to disk.
+
+    The cropped image is saved in WebP format, using the price tag's ID to
+    determine the file path.
+
+    The structure of the path is as follows:
+    /img/price-tags/000/200/000/000200000.webp (for price_tag_id = 200000)
```
raphodn: this means there will be 1 single image per sub-folder, no ?
is that what you want ?

open_prices/proofs/utils.py
```diff
@@ -282,3 +286,58 @@ def match_receipt_item_with_price(receipt_item: ReceiptItem, price: Price) -> bo
...
+
+def get_price_tag_image_path(price_tag_id: int) -> str:
+    """Generate the path for the price tag image based on its ID.
+
+    The path is structured to avoid too many files in a single directory.
+    The ID is zero-padded to 9 digits and split into three parts of three
+    digits each. The image extension is .webp.
+
+    Example: for price_tag_id = 200000, the path will be:
+    /img/price-tags/000/200/000/000200000.webp
```
raphodn: https://prices.openfoodfacts.org/img/price-tags/000/200/000/000200000.webp returns a 404, is just an example or did you share a valid url ?

open_prices/proofs/models.py
raphodn: similar to existing `Proof.file_path` & `Proof.image_thumb_path`
it could be
```
image_path = models.CharField(verbose_name="Proof image cropped using the bounding_box", blank=True, null=True)
```

open_prices/proofs/utils.py
raphael0202: Yes! Once loaded as a Pillow image, it's format-independent.

open_prices/proofs/utils.py
raphael0202: Good catch! There should be only two parts (`000` and `200`), so that we store up to 1000 images per folder.
I'm pushing a fix.

open_prices/proofs/utils.py
raphael0202: It was indeed only an example (and it's not deployed in production yet).

open_prices/proofs/models.py
raphael0202: As the path can be derived directly from the ID, I would avoid storing it in DB, but we can indeed add the field to the serializer to return it in the API indeed!

raphael0202: I saw in Django REST framework doc that it was possible to generate field values on the fly with a method, using `file_path = serializers.SerializerMethodField()`, would it be a good way to implement it? (I'm still new at DRF)

raphodn: I can do it in another PR if you prefer ?
Indeed you can use a SerializerMethodField (ex in the Moderation app/serializer).
Or even via a model property + read_only CharField (see Challenge app/serializer)

raphael0202: Yes, feel free to open a new PR for this!

raphodn: wait actually a serializer-only field is not going to be enough actually..
because it's easy to generate the file path indeed.
but storing the info is the only simple way that tells me if the file exists or not.. and avoid 404s

raphodn: opened a PR here with the serializer-only solution #1143

PR #1139, opened 2025-11-27
chore(format): reformat some files with black

raphael0202:
Some files were not formatted properly.


raphodn:
btw our `pre-commit-config.yaml` file has 3 checks: black, flake8 & isort --> we could put ruff instead
edit: created an issue: #1140

Issue #1112, opened 2025-11-06
Performance issues when many proofs are uploaded at the same time


raphodn:
### Story

The processing queue can get long if there is many PRICE_TAG proofs uploaded at the same time.
We run a few models in cascade : classification, price_tag detection, price_tag extraction...

### Todo

- [ ] increase the number of workers from 2 to 4
- [ ] more RAM ?

PR #1142, opened 2025-11-29
perf: Optimize concurrent proof processing and scale workers

Kinjal0007:
This PR addresses issue #1112, which reports performance bottlenecks when multiple PRICE_TAG proofs are uploaded concurrently.

### 1. Async Optimization
- Implemented `asyncio.Semaphore` in `extract_from_price_tag_batch` to limit concurrent extraction to 5 tasks.
- Prevents overload on remote services and avoids task failures under high concurrency.
- Ensures stable async execution while still improving overall throughput.

### 2. Instrumentation
- Added timing logs inside `ml.py` around batch extraction and model inference stages.
- Provides clearer visibility into where time is spent during bulk proof processing.

### 3. Configuration Improvements
- Updated `settings.py` to allow worker scaling through the `Q_CLUSTER_WORKERS` environment variable.
- Enables maintainers to adjust worker parallelism based on available CPU resources.

## Verification

- Reproduced the issue locally by simulating 30 concurrent uploads.
- Observed bottlenecks and queue saturation in the baseline scenario.
- Verified that:
  - async extraction runs reliably,
  - throughput increases with additional workers,
  - no rate-limit or network-related errors occur,
  - existing API behavior remains correct.

## Deployment Notes

For production environments, the following settings are recommended:

PRICE_TAG_EXTRACTION_ASYNC_REQUESTS=True
Q_CLUSTER_WORKERS=4
GUNICORN_WORKERS=4

## Issue Reference  #1112

### DIFF
--- a/open_prices/proofs/ml.py
+++ b/open_prices/proofs/ml.py
@@ -639,17 +639,28 @@ async def extract_from_price_tag_batch(
         tokens. 0 is DISABLED. -1 is AUTOMATIC.
     :return: a list of Gemini responses, one for each image
     """
+    start_time = time.time()
+    logger.info(f"Starting batch extraction for {len(images)} images")
+    
     responses = []
     # We get the uncached version of the client because otherwise
     # the event loop configuration done during the client initialization
     # leads to an exception when reusing the client in an async context
     # (due to the event loop being closed)
     client = common_google.get_genai_client.__wrapped__()
-    tasks = [
-        extract_from_price_tag_async(client, image, thinking_budget=thinking_budget)
-        for image in images
-    ]
+    
+    # Limit concurrency to avoid rate limits or network errors
+    semaphore = asyncio.Semaphore(5)
+
+    async def sem_task(image):
+        async with semaphore:
+            return await extract_from_price_tag_async(client, image, thinking_budget=thinking_budget)
+
+    tasks = [sem_task(image) for image in images]
     responses = await asyncio.gather(*tasks)
+    
+    duration = time.time() - start_time
+    logger.info(f"Batch extraction finished in {duration:.2f}s")
     return responses
 
 
@@ -1286,12 +1297,19 @@ def run_and_save_proof_prediction(
         return None
 
     image = Image.open(file_path_full)
+    
+    start_time = time.time()
+    logger.info(f"Starting ML predictions for proof {proof.id}")
+
     run_and_save_proof_type_prediction(image, proof)
     run_and_save_price_tag_detection(
         image, proof, run_extraction=run_price_tag_extraction
     )
     if run_receipt_extraction:
         run_and_save_receipt_extraction_prediction(image, proof)
+
+    duration = time.time() - start_time
+    logger.info(f"Finished ML predictions for proof {proof.id} in {duration:.2f}s")
 
 
 def price_tag_prediction_has_predicted_barcode_valid(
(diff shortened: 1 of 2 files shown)


open_prices/proofs/ml.py
```diff
@@ -594,17 +639,28 @@ async def extract_from_price_tag_batch(
...
     # leads to an exception when reusing the client in an async context
     # (due to the event loop being closed)
     client = common_google.get_genai_client.__wrapped__()
-    tasks = [
-        extract_from_price_tag_async(client, image, thinking_budget=thinking_budget)
-        for image in images
-    ]
+    
+    # Limit concurrency to avoid rate limits or network errors
+    semaphore = asyncio.Semaphore(5)
```
raphael0202, line 653: I would make the semaphore count value a function parameter (something called `max_concurrency`).

open_prices/proofs/ml.py
```diff
@@ -594,17 +639,28 @@ async def extract_from_price_tag_batch(
...
+
+    async def sem_task(image):
```
raphael0202, line 655: I don't find it very elegant to use an inlined async func here. It would be better to switch completely to `TaskGroup`s, something like:

```python
async with asyncio.Semaphore(max_concurrency):
    async with asyncio.TaskGroup() as task_group:
        for image in images:
            task_group.create_task(extract_from_price_tag_async(client, image, thinking_budget=thinking_budget))
```

raphael0202 (changes requested):
Thank you for your PR! I've added a few suggestions.

open_prices/proofs/ml.py
raphodn: Can we talk about this on thursday ? I don't really see how this PR helps our current priorities, except complexifying/over-engineering ?

raphodn: @Kinjal0007 it's the 2nd PR you open without challenging the issue or trying to understand the needs first (see #1124).
Could you start by putting a comment in the issues ? and explaining your solution ?

raphodn: And we need to improve/clean the issues on our side and better pinpoint the "Good first issues" :pray:

raphael0202: Yes, we can talk about this on Thursday!
Even though the current implementation won't solve the issue, it's still good to limit the number of concurrent async requests made. On the network issue with Gemini, I was thinking of adding a second implementation using another LLM framework (such as https://ai.pydantic.dev), to see whether the issue we face comes from the `google-genai` SDK (which I suspect).

Kinjal0007: @raphodn @raphael0202 

Thanks for the feedback

I understand the concern about adding complexity without aligning on priorities first, apologies .

Kinjal0007: @raphael0202 

should I approach the new suggestion or should I  drop it here ?

raphael0202: We have a weekly meeting every Thursday over Google Meet, if you're available you're welcome to join to discuss it!
Otherwise, we will update the issue after talking about it during the meeting.

raphael0202:
Since we merged #1163, we don't have network issues with Gemini anymore, closing the client was the right approach!
I'm closing this PR, as having too many concurrent requests was not the culprit here.

PR #1143, opened 2025-12-01
refactor(Proofs): API: return PriceTag cropped image_path

raphodn:
### What

Following #1138
Return the `image_path` in the PriceTag list API
Using 2 new model properties : image_path & image_path_full

### DIFF
--- a/open_prices/proofs/models.py
+++ b/open_prices/proofs/models.py
@@ -566,17 +566,13 @@ class PriceTag(models.Model):
 
     @property
     def image_path(self):
-        from open_prices.proofs.utils import get_price_tag_image_path_full
+        from open_prices.proofs.utils import get_price_tag_image_path
 
-        return get_price_tag_image_path_full(self.id)
-
-    @property
-    def image_path_display(self):
-        return self.image_path if os.path.exists(self.image_path) else None
+        return get_price_tag_image_path(self.id)
 
     @property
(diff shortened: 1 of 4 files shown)


open_prices/proofs/models.py
```diff
@@ -563,6 +564,13 @@ def save(self, *args, **kwargs):
         self.full_clean()
         super().save(*args, **kwargs)
 
+    @property
+    def image_path(self):
+        from open_prices.proofs.utils import get_price_tag_image_path
+
+        image_path = get_price_tag_image_path(self.id)
+        return image_path if os.path.exists(image_path) else None
```
raphodn: not a fan of having to import + check if the file exists + ternary for each returned PriceTag 👀 
(vs storing the info in a dedicated model field)

raphael0202: Personally, I wouldn't check at all: just return the image path. In case it is missing server-side, a 404 status code will be returned by the nginx server, which can be handled gracefully client-side.
And I'm not a fan of the overhead of checking that all image exists when you fetch proofs in bulk (such as on the proof page).

raphael0202 (approved):

open_prices/proofs/models.py
raphodn: ok i'll return the raw path :+1:

PR #1165, opened 2025-12-21
fix(PriceTag): actually store barcode normalization (post-ML)

### LINKED ISSUE #1080: Normalize barcode after Gemini price tag extraction
Predicted barcodes are currently not normalized. They are when the price is saved (see #1044), but not normalizing before prevents to display the product even when it exists on OFF.

cf https://openfoodfacts.slack.com/archives/C08AQU2C1NK/p1760651312901129?thread_ts=1760005408.129539&cid=C08AQU2C1NK

### Screenshot

Image

raphodn:
### What

fix of #1085. We added barcode normalization, but weren't actually storing the result.
Same, we had a barcode_raw but the `LabelWithSimilarBarcodes` model didn't have the key defined.

Closes #1080

### DIFF
--- a/open_prices/proofs/ml/price_tags.py
+++ b/open_prices/proofs/ml/price_tags.py
@@ -504,7 +504,7 @@ class LabelWithSimilarBarcodes(Label):
     are sorted by increasing Levenshtein distance.
     """
 
-    barcode_raw: str = Field(
+    raw_barcode: str = Field(
         ..., description="The raw extracted barcode, before normalization."
     )
     similar_barcodes: list[BarcodeSimilarityMatch] = Field(
@@ -702,7 +702,7 @@ def run_and_save_price_tag_extraction(
         # 1) fix barcode with some custom rules
         # 2) if the barcode is still unknown, generate similar barcodes
         barcode = response.parsed.barcode
-        barcode_raw = barcode
+        raw_barcode = barcode
         similar_barcodes = []
         # 1) barcode fix
         if barcode:
@@ -734,7 +734,7 @@ def run_and_save_price_tag_extraction(
 
         data = LabelWithSimilarBarcodes(
             **{**dict(response.parsed), "barcode": barcode},  # merge any barcode fix
-            barcode_raw=barcode_raw,
+            raw_barcode=raw_barcode,
             similar_barcodes=similar_barcodes,
         )
         try:


open_prices/proofs/ml/price_tags.py
```diff
@@ -504,6 +504,9 @@ class LabelWithSimilarBarcodes(Label):
     are sorted by increasing Levenshtein distance.
     """
 
+    barcode_raw: str = Field(
```
raphael0202: I think `raw_barcode` would be a better name (and it would be better as well to rename the variable in the func)

open_prices/proofs/ml/price_tags.py
```diff
@@ -728,8 +731,9 @@ def run_and_save_price_tag_extraction(
                 # Check that barcode is valid (correct check digit)
                 if common_openfoodfacts.barcode_is_valid(p.code)
             ]
+
         data = LabelWithSimilarBarcodes(
-            **dict(response.parsed),
+            **{**dict(response.parsed), "barcode": barcode},  # merge any barcode fix
```
raphael0202, line 736: You're not saving `barcode_raw` here, no?

raphodn: In the line below ?

open_prices/proofs/ml/price_tags.py
raphodn: Hmm i'm more of the naming school that suffixes rather than prefixes ^^
But i can make the change

open_prices/proofs/ml/price_tags.py
raphael0202: I mean `LabelWithSimilarBarcodes` is not the schema provided to Gemini, it's a class derived from `Label` (which is the schema passed to Gemini), with additional fields.

open_prices/proofs/ml/price_tags.py
raphael0202: So far we used mostly (only?) prefixes. But it's a debug field, so I'm fine with leaving it as-it :)

open_prices/proofs/ml/price_tags.py
raphodn: Well barcode_raw was already set in the line below, but the class didn't have it, so when data was instantiated it disappeared.

raphael0202: Oh, I missed the line below 😅 ! Sorry for that!
It's all good for me.

Issue #1151, opened 2025-12-14
Django: upgrade to 5.2


raphodn:
### Story

We currently have Django 5.1.X

https://docs.djangoproject.com/en/5.2/releases/5.2/

AnasDharar:
Hello @raphodn sir, I have submitted a PR for this issue
Would appreciate a review when you have time

PR #1178, opened 2025-12-31
chore: upgrade Django to 5.2

shubhamkoti:
## Summary
This PR upgrades Django from 5.1.x to 5.2.

## Changes
- Updated Django dependency to 5.2
- Regenerated `poetry.lock`
- Guarded optional dependencies (e.g. Sentry, CORS) to allow local Django checks when tooling is not installed

## Notes
Local installation on Windows fails due to a Linux-only transitive dependency (`perf-analyzer`), but CI/Linux environments are unaffected.

## References
- Django 5.2 release notes: https://docs.djangoproject.com/en/5.2/releases/5.2/
- Issue: #1151

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -37,7 +37,8 @@ DJANGO_APPS = [
     "django.contrib.staticfiles",
 ]
 
-THIRD_PARTY_APPS = [ 
+THIRD_PARTY_APPS = [
+    "corsheaders",  # django-cors-headers
     "rest_framework",  # djangorestframework
     "django_filters",  # django-filter
     "drf_spectacular",  # drf-spectacular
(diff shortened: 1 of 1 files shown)


config/settings.py
raphodn, line 1: Hi ! I don't understand the changes in this file. Is this breaking for Django 5.2 ? It seemed to work well for 5.1.X.

shubhamkoti: Thanks for pointing this out!

You’re right — the changes in `config/settings.py` are not required for the Django 5.2 upgrade itself. They were added to make local execution easier on my environment, but they are unrelated to Django 5.2 specifically.

To keep this PR focused on the upgrade, I’ll revert the `settings.py` changes and limit the PR to:
- upgrading Django to 5.2
- updating the lockfile accordingly

Thanks for the review!

PR #1181, opened 2026-01-05
fix(scheduler): increase timeout for specific task

raphael0202:
Increase timeout for update_product_counts_task.
This task takes more than 6 hours to complete now.

### DIFF
--- a/open_prices/common/tasks.py
+++ b/open_prices/common/tasks.py
@@ -159,7 +159,7 @@ CRON_SCHEDULES = {
     ),
     "update_product_counts_task": (
         "20 2 * * 1",  # every start of the week (at 02:20)
-        {"timeout": 6 * 60 * 60},  # 6 hours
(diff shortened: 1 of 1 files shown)


open_prices/common/tasks.py
```diff
@@ -159,7 +159,7 @@ def history_cleanup_task():
     ),
     "update_product_counts_task": (
         "20 2 * * 1",  # every start of the week (at 02:20)
-        {"timeout": 6 * 60 * 60},  # 6 hours
+        {"timeout": 10 * 60 * 60},  # 10 hours
```
raphodn, line 162: we'll also need to look into reducing the task's effective time ^^

raphael0202: I already looked into it, there may be some optimizations to be done, but there is no obvious shortcut here, we're updating 4M rows

PR #1184, opened 2026-01-08
feat(Prices): add `duplicate_of` FK field

### LINKED ISSUE #422: Detect duplicate prices
### Story

I sometimes see "identical" prices. Reasons could be a mistake (scanning/adding the product twice), or adding a price from a receipt and then from another source like GDPR, API error...)

Identical price = same product + same location + same date
(not necessarily the same proof ; if 2 different proofs, which one to keep ?)

raphael0202:
- add `prices.duplicate_of_id` field, that points to the reference price
- add a update_duplicate_of_field command to refresh to `duplicate_of_id` field for all prices
- when a price is deleted, run again duplicate search for prices pointing to the price

Closes #422.

Notes:
- I had to comment `migrations.RunPython(populate_history),` in `open_prices/prices/migrations/0013_historicalprice.py`, otherwise running migrations (before unit tests) failed (after I updated the Prices table).

@raphodn I would like your opinion on this, but I think we should drop the populate history function call in the migration script. The unit test is failing too on the CI: https://github.com/openfoodfacts/open-prices/actions/runs/20823717246/job/59819147762?pr=1184

### DIFF
--- a/open_prices/api/prices/filters.py
+++ b/open_prices/api/prices/filters.py
@@ -79,5 +79,4 @@ class PriceFilter(django_filters.FilterSet):
             "proof_id": ["exact", "in", "isnull"],
             "owner": ["exact"],
             "created": ["gte", "lte"],
-            "duplicate_of_id": ["isnull"],
         }


raphael0202:
@raphodn Do you have some time to review this PR? :)

open_prices/prices/models.py
```diff
@@ -329,6 +339,7 @@ def save(self, *args, **kwargs):
         # self.set_proof()  # should already exist
         self.set_product()
         self.set_location()
+        self.set_is_duplicate_of()
```
raphodn, line 342: so on each save (creation, but also edition), it'll run an extra SQL query in sync..

raphodn: i think i'd would have gone with an async solution
- on post_save in a task
- or nightly task looking at the prices created during the day

raphodn: sure there's already SQL queries above with product & location. 
do we need to return the info "this price is a duplicate" immediately ? if yes then indeed there's not that many other solutions 🤔

open_prices/prices/models.py
```diff
@@ -392,6 +403,43 @@ def in_challenge(self, challenge: Challenge):
...
+          price information (price, discount type, is discounted, price
+          without discount), product code/category tag, labels tags, and
+          origins tags
+        - it was created later than the other price (we keep the first one)
+        """
+        # Look for possible duplicates
+        possible_duplicates = (
+            Price.objects.filter(
+                type=self.type,
+                location_osm_id=self.location_osm_id,
```
raphodn: what about using location_id instead ?
to manage ONLINE prices as well
the location_id should be set as we call `set_location` just before

raphodn: minor risk : prices being added without a location and/or date and/or etc...
it's possible i believe, as prices should share their proof info, and we're quite flexible regarding the proof fields.
so maybe add some additional `{field}__isnull=False` in the queryset ?

open_prices/prices/models.py
```diff
@@ -392,6 +403,43 @@ def in_challenge(self, challenge: Challenge):
...
+                price_is_discounted=self.price_is_discounted,
+                price_without_discount=self.price_without_discount,
+                discount_type=self.discount_type,
+                product_code=self.product_code,
+                category_tag=self.category_tag,
+                labels_tags=self.labels_tags,
+                origins_tags=self.origins_tags,
+            )
+            .exclude(id=self.id)
+            .order_by("created")
```
raphodn: could `id` be a sufficient proxy here ?
the Price table is already ordered by id, there's no additional ordering.
so i'd say this line is not useful (and avoids an SQL ORDER BY on a non-indexed column)

open_prices/prices/models.py
```diff
@@ -456,3 +504,16 @@ def price_post_delete_decrement_counts(sender, instance, **kwargs):
         Location.objects.filter(id=instance.location_id, price_count__gt=0).update(
             price_count=F("price_count") - 1
         )
+
+
+@receiver(signals.post_delete, sender=Price)
```
raphodn, line 516: I believe `on_delete=models.SET_NULL` of the ForeignKey does the job :)

```
1. price A is added
2. price B is added, and is a duplicate of A : B.duplicate_of_id = A.id (and A.duplicates = [B])
3. if you delete A : B still exists, and duplicate_of_id is set to None
```

```
1. price A is added
2. price B is added, and is a duplicate of A : duplicate_of_id = A.id
3. if you delete B : A still exists, the related_name returns A.duplicates = []
```

raphodn: or is this in case of 2 or more duplicates of the same price ?

Makefile
```diff
@@ -126,7 +126,7 @@ django-tests:
 
 django-tests-single: guard-args
 	@echo "🥫 Running specific tests …"
-	${DOCKER_COMPOSE_TEST} run --rm api poetry run python3 manage.py test -v 2 ${args}
+	${DOCKER_COMPOSE_TEST} run -e 'Q2_SYNC=True' --rm api poetry run python3 manage.py test -v 2 ${args}
```
raphodn: other PR ? 😁

open_prices/prices/management/commands/update_duplicate_of_field.py
```diff
@@ -0,0 +1,17 @@
...
+
+class Command(BaseCommand):
+    """Refresh the `duplicate_of_id` field for all prices, by saving them
+    trigger any updates."""
+
+    help = "Refresh the `duplicate_of_id` field for all prices."
+
+    def handle(self, *args, **options) -> None:
+        self.stdout.write("Updating duplicate_of field for all prices...")
+        for price in tqdm.tqdm(Price.objects.all(), desc="prices"):
```
raphodn: i wonder if we could do this in 1 single SQL query instead.
at least to isolate the duplicate prices. and then loop on them to set duplicate_of + save + historical entry (whereas doing a bulk query/update would skip the save + historical entry)

open_prices/prices/models.py
raphael0202: > so on each save (creation, but also edition), it'll run an extra SQL query in sync..

That's the idea, I would favor having this information in real-time, especiallly if by default we hide these prices on the front-end. The query is quite lightweight, as it's just a SELECT with many filters (many of them with index).

open_prices/prices/models.py
raphael0202: Yes, good idea to order by ID. I would keep the line though, even if it's the default, to make it transparent to the reader why we need price ordered by creation date.

open_prices/prices/models.py
raphael0202: > or is this in case of 2 or more duplicates of the same price ?

Yes exactly, if the reference price (that all other duplicate prices refer to) is deleted, we need to find a new reference price and set the `duplicate_of_id` field accordingly for other prices.

open_prices/prices/management/commands/update_duplicate_of_field.py
raphael0202: To be honest, I wouldn't spend much time on this. The query is quick to launch, and we should only launch it once to populate the `is_duplicate_of_id` field, after this the `set_is_duplicate_of` method takes care of keeping the field up to date.

raphodn: i doubt it.. you run `.save()` on all 150k prices.
that means validations will run for each of them.
there will likely be errors on some of them + it will update their `updated` field even if no other field was changed

open_prices/prices/models.py
raphael0202: I now filter on `location_id` instead.

raphael0202: I also now check that `date` and `location_id` fields are not null.

open_prices/prices/management/commands/update_duplicate_of_field.py
raphael0202: I didn't think of the `updated` field. I've pushed a new version where we first group prices by duplicate, then iterate over the prices.

raphael0202: Otherwise, I wouldn't mind that the script takes a few hours to run, as it's supposed to be launched only once :)
How come we could hit validation error on existing prices in the DB?

raphodn: the bigger the database grows, the likelier we get strange data ^^
i do nightmares of this at ma-cantine :laughing: 
but if you're confident enough.. :)

open_prices/api/prices/filters.py
```diff
@@ -79,4 +79,5 @@ class Meta:
             "proof_id": ["exact", "in", "isnull"],
             "owner": ["exact"],
             "created": ["gte", "lte"],
+            "duplicate_of_id": ["isnull"],
```
raphodn: moving this to another PR so that we can test it seperately

raphodn: done in #1204

open_prices/prices/management/commands/update_duplicate_of_field.py
raphael0202: Indeed, that's a valid point ;)
Having a script to perform sanity checks to detect validation issues would be interesting though, to be confident that this kind of script wouldn't fail in production!

PR #715, opened 2025-02-21
feat(Proofs): ML: receipt content extraction

TTalex:
### What
- Uses gemini to extract receipt content
- Stores it in the proof_predictions database

### Notes

I did try to ask Gemini for way more information (see commented fields), but it tends to confuse it. Sometimes it replies with all fields, and sometimes it forgets to add everything, including price.
```python
class Item(typing.TypedDict):
    product: Products
    price: float
    # origin: Origin
    # unit: Unit
    # organic: bool
    # barcode: str
    product_name: str
    # is_discount: bool
    # discount_price: float

class Receipt(typing.TypedDict):
    store_name: str
    store_address: str
    store_city_name: str
    date: str
    # number_of_articles: int
    # total_price: float
    items: list[Item]
```

The shorter list seems to perform better.

Extraction of store name, address and date could be quite useful. However, in the current process, these are already specified by the user when the proof reaches the ml (also true for proof type prediction).

Since users currently manually mark the proof as a receipt, we use this info to only run receipt extraction on such proofs. We could consider using the proof type prediction instead.

Finally, this whole process requires the `ENABLE_ML_PREDICTIONS` env to be True because it is called at the same time as other proof ml processes. Some tiny changes are required to run this locally if no Triton Inference Server is running. The Triton server isn't required for this process, but is mandatory for the other ones.

### Fixes bug(s)
- Fixes #681 

### Screenshot
Some examples of extraction:

(Note the confusion with the "PFAND", the money-back german system when you take or return glassware)
```json
{
    "date": "",
    "items": [
        {
            "price": 0.65,
            "product": "en:bananas",
            "product_name": "FRUCHTJ. VAN&BALL"
        },
        {
            "price": 1.29,
            "product": "en:cherries",
            "product_name": "KIRSCH GRUETZE"
        },
        {
            "price": 3.38,
            "product": "en:mandarin-oranges",
            "product_name": "JARRITOS MANDARI"
        },
        {
            "price": 0.5,
            "product": "other",
            "product_name": "PFAND 0,25 EURO"
        },
        {
            "price": 1.09,
            "product": "en:strawberries",
            "product_name": "STRAWB. KIWI"
        },
        {
            "price": 0.25,
            "product": "other",
            "product_name": "PFAND 0,25 EURO"
        }
    ],
    "store_address": "Reichenhainer Str. 55",
    "store_city_name": "Chemnitz",
    "store_name": "Nahkauf Oliver Kettler"
}
```

```json
{
[...]

PR #1185, opened 2026-01-09
feat: improve receipt extraction (schema v2)

raphael0202:
This PR improves the extraction schema for receipts (v2 of #715). We add a schema version, as for price tag schema.

The most notable additions are:
- `item[].type`: we now predict the type of the price directly. This allows to select the right type for the entry in the receipt assistant
- `item[].category_group`: we extract the "category group" of the entry, if any. This is useful for some research studies.
- `item[].product_code`: extract the barcode is present, not used yet on the front-end.
- `item[].category` and `item[].category_tag`: product was renamed in category_tag, and category is a free-text field to save the category in English. This way, we can find a more precise mapping manually, and fallback to category_tag if not found (to be done in a future PR)
- `item[].price_total` and `price`: we consider separately the price per unit/kg and the price actually paid (we're not interested in the latter for the moment)
- `item[].price_without_discount` & `price_is_discounted` & `discount_type`: we now support discounted price! It's not supported on the front-end yet, however (there is a bug related to currency when passing the price in the `PriceDiscountChip`)
- `item[].quantity`: we can now parse the quantity for type=CATEGORY! No need to type everything manually anymore.
- more store-related metadata, with better JSON schema descriptions.

### DIFF
--- a/open_prices/proofs/ml/receipts.py
+++ b/open_prices/proofs/ml/receipts.py
@@ -114,7 +114,9 @@ class ReceiptItemType(BaseModel):
         ...,
         description="Unit of the price. Can be one of: KILOGRAM: price is per kg, only if type = CATEGORY; "
         "LITER: price is per liter, only if type = CATEGORY; "
-        "UNIT: price is per unit (available for both CATEGORY and PRODUCT types). Set to null if not detected.",
+        "UNIT: price is per unit (available for both CATEGORY and PRODUCT types). "
+        "If the unit displayed is not a SI unit (ex: ounce, pound), it must be converted to the closest SI unit. "
+        "Set to null if not detected.",
     )
     product_name: str | None = Field(
         ...,


open_prices/proofs/ml/receipts.py
```diff
@@ -1,40 +1,190 @@
...
+class ReceiptItemType(BaseModel):
+    type: Literal["PRODUCT", "CATEGORY"] = Field(
+        ...,
+        description="The type of product the receipt entry is referring to. It should be "
+        "`PRODUCT` for packaged products with barcode, and `CATEGORY` for raw "
+        "products without barcode. The barcode is usually not displayed on receipts, "
+        "so the `CATEGORY` type is often inferred from the product name and from the fact "
+        "the product is sold per weight.",
+    )
+    category_group: str | None = Field(
```
TTalex, line 35: Should this one have an enum type instead of str ? It would make it easier to process later on.
Can we have an enum type and still let the LLM offer other options if we forgot one in the enum ?

open_prices/proofs/ml/receipts.py
```diff
@@ -1,40 +1,190 @@
...
+    )
+    quantity: float | int | None = Field(
+        ...,
+        description="the quantity for the product entry. It is a number, either an integer "
+        "(for packaged goods or for fruits and vegetables that can be bought "
+        "individually), or a float representing the weight (per kg or per liter), for "
+        "fruits, vegetables and raw foods that are sold by weight. Set to "
+        "null if not detected.",
+    )
+    price_per: Unit | None = Field(
```
TTalex, line 113: Makes me wonder what happens when the unit isn't Kg, but pounds or ounces. Does it converts ?
Can't remember what we did on price tags

open_prices/proofs/ml/receipts.py
```diff
@@ -1,40 +1,190 @@
...
+        description="the phone number of the store. It should be null if the "
+        "phone number is not known. Example: '+33123456789'",
+    )
+    date: str | None = Field(
+        ...,
+        description="the date when the receipt was issued, in ISO 8601 format "
+        "(YYYY-MM-DD). It should be null if the date is not known. Example: "
+        "'2023-10-15'",
+    )
+    hour: str | None = Field(
```
TTalex, line 168: Is that necessary with the date field ? ISO8601 includes hours

open_prices/proofs/ml/receipts.py
raphael0202: As the category group depends on the shop type we're in, I think it's difficult to predict the category groups. And as it's more for a research perspective, I prefer letting the reuser classify the category group himself/herself.

open_prices/proofs/ml/receipts.py
raphael0202: That's true, to prevent any ambiguity, we should improve the description to ask to convert if needed!

open_prices/proofs/ml/receipts.py
raphael0202: We're not really interested in the hour in Open Prices, just the date, so extracting the date only means we don't have to parse the field to populate the `date` field.
I only added the `hour` field out of completeness, as some people may find it interesting.

open_prices/proofs/ml/receipts.py
raphael0202: I fixed this

PR #1186, opened 2026-01-11
feat(PriceTags): Allow filtering on proof location

TTalex:
### What
- It's pretty hard to validate prices of products you don't know, in a language you can't speak
- This PR allows filtering of price tags according to their proof's country codes.


open_prices/api/proofs/filters.py
raphodn, line 52: the `PriceTagFilter` is used in the `PriceTagViewSet`, which has a queryset with a `.select_related("proof__location")`, so it should be good in terms of performance 👍

raphodn: also added tests here : https://github.com/openfoodfacts/open-prices/pull/1186/commits/33de72214b1800205c1f9025552abf39359ee3b8

PR #1201, opened 2026-01-22
fix: prevent foreign key violation for deleted proofs

areebahmeddd:
**Signed-off-by**: Areeb Ahmed [areebshariff@acm.org](mailto:areebshariff@acm.org)

## Summary of Changes
- async prediction tasks fail with foreign key violations when proofs are deleted before task execution

## Related Issue(s)
Closes ~~#893~~ #798

### DIFF
--- a/open_prices/proofs/ml/__init__.py
+++ b/open_prices/proofs/ml/__init__.py
@@ -36,14 +36,6 @@ def run_and_save_proof_prediction(
     :param run_price_tag_extraction: whether to run the price tag extraction
         model on the detected price tags, defaults to True
     """
-    try:
-        proof = Proof.objects.get(id=proof.id)
-    except Proof.DoesNotExist:
-        logger.warning(
-            "Proof with id %s no longer exists, skipping ML prediction", proof.id
-        )
-        return None
-
     file_path_full = proof.file_path_full
 
     if file_path_full is None or not Path(file_path_full).exists():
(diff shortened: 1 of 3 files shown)


open_prices/proofs/ml/__init__.py
```diff
@@ -36,6 +36,14 @@ def run_and_save_proof_prediction(
     :param run_price_tag_extraction: whether to run the price tag extraction
         model on the detected price tags, defaults to True
     """
+    try:
```
raphodn: the existing `if` below checks if the proof image actually exists, and stops if not.
so there's kinda already a check if the proof actually exists or not.
which makes me doubt that this PR will actually fix the issue..

the solution is probably to be done further down the line..

raphodn: - run_and_save_proof_type_prediction : the `ProofPrediction.objects.create()` call is wrapped in a try / catch
- run_and_save_price_tag_detection : the `ProofPrediction.objects.create()` call is NOT wrapped in a try / catch
- run_and_save_receipt_extraction_prediction : `ProofPrediction.objects.create()` call is wrapped in a try / catch

raphodn: so i'd rather first add a try / catch in run_and_save_price_tag_detection

areebahmeddd: makes sense! added trycatch in price tag detection matching other functions

PR #1204, opened 2026-01-23
feat(Prices): API: add a filter on the new duplicate_of field

raphodn:
### What

Following #1184


open_prices/prices/models.py
```diff
@@ -418,7 +417,7 @@ def set_is_duplicate_of(self):
         possible_duplicates = (
             Price.objects.filter(
                 type=self.type,
-                location_id=self.location_id,  # type: ignore
```
raphael0202, line 421: The `type: ignore` is used to tell mypy to ignore the error (as `self.location_id` is not a known attribute of the class, it's build dynamically).

PR #1208, opened 2026-01-26
feat: approximate counting with pg_class.reltuples

### LINKED ISSUE #948: Counting the number of products in the products table is slow
Slow DB Query

Sentry Issue: [OPEN-PRICES-22](https://openfoodfacts.sentry.io/issues/5821359823/?referrer=github_integration)

```sql
SELECT COUNT(*) AS __count
FROM products
```

areebahmeddd:
**Signed-off-by**: Areeb Ahmed [areebshariff@acm.org](mailto:areebshariff@acm.org)

## Summary of Changes
- custom query set that uses `reltuples` for unfiltered counts
- falls back to exact `count (*)` when filters are applied

ref: https://testdriven.io/blog/django-approximate-counting

## Related Issue(s)
Closes #948

### DIFF
--- a/open_prices/common/managers.py
+++ b/open_prices/common/managers.py
@@ -3,6 +3,16 @@ from django.db.models import Manager, QuerySet
 
 
 class ApproximateCountQuerySet(QuerySet):
+    """Use PostgreSQL's pg_class.reltuples for fast approximate counting.
+
+    For large tables, COUNT(*) requires a full table scan which is slow.
+    This queryset uses PostgreSQL's internal table statistics (reltuples)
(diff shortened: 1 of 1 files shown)


open_prices/common/managers.py
```diff
@@ -0,0 +1,24 @@
+from django.db import connections
+from django.db.models import Manager, QuerySet
+
+
+class ApproximateCountQuerySet(QuerySet):
```
raphael0202, line 5: Can you add a docstring to explain the interest of the class, and why we're implementing this?

PR #1220, opened 2026-02-09
fix: handle empty strings in CSRF_TRUSTED_ORIGINS and untrack .env

keshavagarwal321:
## Description
This PR addresses two configuration issues:
1. **Fragile CSRF Parsing:** The `CSRF_TRUSTED_ORIGINS` parsing logic would crash with a `SystemCheckError` if the environment variable contained trailing commas or empty strings (e.g., `http://localhost:8000,`). I added a list comprehension filter to strip empty strings.
2. **Security:** The `.env` file was being tracked by git. This is a security risk as it often contains local secrets. I have removed it from tracking and updated `.gitignore` to prevent future commits of this file.

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Security fix

## How Has This Been Tested?
- Verified locally that the server starts successfully even when `CSRF_TRUSTED_ORIGINS` contains trailing commas.
- Verified that `git status` no longer tracks `.env`.

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,84 @@
+# unify separator with windows style
+COMPOSE_PATH_SEPARATOR=;
+# dev is default target
+COMPOSE_FILE=docker-compose.yml;docker/dev.yml
+
+# for dev only: connection to local product opener network and for Product Opener redis
+# in staging
+COMMON_NET_NAME=po_default
+
+SECRET_KEY=key
+
+DEBUG=True
+
+# Set to True if you want to run Django Q2 tasks in sync mode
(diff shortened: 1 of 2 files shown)


.env
areebahmeddd, line 1: this .env file is required by docker compose! don’t delete it. it’s safe for git since it doesn’t contain sensitive info

keshavagarwal321: My apologies! I was following standard security practices and didn't realize that this project uses .env for Docker defaults. I've reverted the removal. Thanks for the review!

PR #1225, opened 2026-02-12
chore: switch from poetry to uv

### LINKED ISSUE #1221: Use uv instead of poetry
### Story

https://docs.astral.sh/uv/

@raphael0202 I saw you did recently here https://github.com/openfoodfacts/openfoodfacts-python/pull/431 😁 

### Why

- faster
- easier to setup
- we already use ruff (#1150)

raphodn:
### What

Closes #1221


pyproject.toml
```diff
@@ -18,60 +18,62 @@ module = [
...
 
-[tool.poetry]
+[project]
 name = "open-prices"
 version = "0.1.0"
 description = "An open database of product prices"
-authors = ["Open Food Facts <contact@openfoodfacts.org>"]
-license = "AGPL-3.0 licence"
 readme = "README.md"
+requires-python = ">=3.11,<3.12"
```
raphodn: i'm not too sure about this line

Dockerfile
raphodn, line 1: the tests seem to fail, likely the docker config ?

pyproject.toml
raphael0202: It should be compatible with python>=3.12, you can just have `requires-python = ">=3.11"
`

Dockerfile
raphael0202: The uv doc gives some advice about Docker integration: https://docs.astral.sh/uv/guides/integration/docker/

You can directly copy uv and uvx from their Docker image into your own image

pyproject.toml
raphodn: ok fixed here: https://github.com/openfoodfacts/open-prices/pull/1225/commits/05e20bc998b633b7052c9d2f3885c9bfc55ee9f4

though the code has "3.11" hardcoded in many places.
i think i had an issue when setting up the project yesterday, i had 3.12 and it didn't work (forgot where, maybe pre-commit)

Issue #866, opened 2025-07-03
Authentication: use Open Food Facts' keycloak


raphodn:
### Story

TBD

### Related info

- https://github.com/openfoodfacts/openfoodfacts-server/pull/9251

teolemon:
cc @john-gom, maybe a small meeting with @raphodn

PR #1235, opened 2026-02-21
feat(Auth): Support for keycloak tokens

TTalex:
### What
- Updated the /auth API endpoint to also support access_tokens generated by a Single Sign On using keycloak on the frontend
- To minimize impact on codebase, username/password flow is still supported, and Session model is still used

**Here's the full end-to-end flow**
1. Frontend: User goes to the Sign In page
2. Frontend: User presses the Sign In with Open Food Fact button
3. Frontend: User is redirected to auth login form (hosted on keycloak)
4. Frontend: User enters username and password if they weren't already logged in on keycloak
5. Frontend: User gets redirected to the Sign In page, with tokens information (id_token, refresh_token and access_token)
   * id_token is usually used to check user identity and should be used by the frontend
   * refresh_token is usually used to fetch new tokens, since they expire quickly (unused for now)
   * access_token is usually used to access resources, OpenId Connect is pretty permissive on its content, OpenFoodFacts' Keycloak includes some user info inside. Thus this is the token we'll send back to the Open Prices backend
7. Frontend: access_token is used to call the /api/auth API
8. Backend: access_token is extracted from request, after param validation
    * Since we also use the existing /api/auth endpoint, a new validation rule was added : queries should either provide username/password OR access_token. Open to suggestions on this one, it's a bit clunky.
9. Backend: In order to decode the token and check its authenticity, server config params are fetched from the keycloak server (using a .well-know endpoint, such as https://auth.openfoodfacts.org/realms/openfoodfacts/.well-known/openid-configuration)
    * This endpoint should be configured in .env `KEYCLOAK_OIDC_CONFIG_URL`
10. Backend: These params also contain a url used to fetch the public keys needed to verify token signature. This is done to make sure that the expected server has issued the token, since the server has signed the token using the corresponding private key.
11. Backend: Some claims are also checked at this point: issuer and audience. Issuer is another check on auth server identity. Audience makes sure the token was issued for this app, and not another one.
    * The audience param should be configured in .env `KEYCLOAK_AUDIENCE`. Example: it's set to 'account' in the public_test_client configured on [openfoodfacts-auth](https://github.com/openfoodfacts/openfoodfacts-auth)
12. Backend: We now trust the token, and it has been decoded, we extract the `preferred_username` field and use it to create or fetch a user/session. 
    * Note: this is not a default OpenId Connect field, it is specific to the OpenFoodFacts Keycloak instance to easily access OFF username
14. Backend: User info is sent back to frontend. This is the same reply as the previous username/password flow
15. Frontend: Response data is saved in the app store. User is now logged in, and redirected to their dashboard

**Some tools and references**
* https://www.jwt.io/ : Great to decode tokens and check the expected content. For example, we can see the expected audience (`aud`)
* https://pyjwt.readthedocs.io/en/stable/usage.html#retrieve-rsa-signing-keys-from-a-jwks-endpoint : Documentation of the new pyjwt dependency, paragraph on the signing key validation
* https://www.keycloak.org/securing-apps/javascript-adapter : Keycloak documentation, paragraph on the javascript adapter new dependency used in frontend
* https://github.com/openfoodfacts/openfoodfacts-auth : Keycloak configuration for Open Food Facts. Needed to set up a local keycloak instance with proper configs. The readme also has pretty good documentation on the whole auth process

**Config required on Keycloak Off instances**
* A public client needs to be configured for Open Prices to use. We could in theory use the same one as OpenFoodFacts, but it's usually not a best practice.
* The following params should be required to create a Keycloak public client:
    * Valid redirect URIs : https://prices.openfoodfacts.org/sign-in (.net for staging)
    * Web origins : https://prices.openfoodfacts.org/ (.net for staging)
    * Auth flow: Standard
* In return, the keycloak administrator should give us:
    * Auth server url : should be https://auth.openfoodfacts.org/realms/openfoodfacts  (.net for staging) 
        * -> To add to frontend env `VITE_KEYCLOAK_URL`
    * Oidc config url : should be https://auth.openfoodfacts.org/realms/openfoodfacts/.well-known/openid-configuration 
        * -> To add to backend env `KEYCLOAK_OIDC_CONFIG_URL`
    * Realm : should be openfoodfacts (also part of the urls above) 
        * -> To add to frontend env `VITE_KEYCLOAK_REALM `
    * Client Id 
        * -> To add to frontend env `VITE_KEYCLOAK_CLIENT_ID`
    * Access token audience : should be "account", can sometimes be the same as the client id  
        * -> To add to backend env `KEYCLOAK_AUDIENCE`

In the future, we might also need a private client in order to auth queries to Product Opener on behalf of users (production creation / edition). The process will most likely require a token exchange, using the access_token generated using the public client, and exchanging it for another access_token valid for API calls (private client).
This might require some changes, to work around token expiration. But this is out of scope of this PR.

### Demo video (frontend)

https://github.com/user-attachments/assets/2da07289-710c-4397-947d-e5078e794d06

Shown in the video:
* Log in using OFF Keycloak, using username / password
* Edits still working
* Log out on Open Prices (doesn't log out the keycloak session)
* Log in using OFF Keycloak, skipping username / password since keycloak session is still active

### Fixes bug(s)
- Fixes https://github.com/openfoodfacts/open-prices/issues/866

### Related frontend PR
- https://github.com/openfoodfacts/open-prices-frontend/pull/2039

### DIFF
--- a/.github/workflows/container-deploy.yml
+++ b/.github/workflows/container-deploy.yml
@@ -44,6 +44,8 @@ jobs:
         # so we use the production redis server
         echo "REDIS_HOST=10.1.0.113" >> $GITHUB_ENV
         echo "REDIS_PORT=6379" >> $GITHUB_ENV
+        echo "KEYCLOAK_OIDC_CONFIG_URL=http://auth.openfoodfacts.net/realms/openfoodfacts/.well-known/openid-configuration" >> $GITHUB_ENV
+        echo "KEYCLOAK_AUDIENCE=account" >> $GITHUB_ENV
     - name: Set various variable for production deployment
       if: matrix.env == 'open-prices-org'
       run: |
@@ -57,6 +59,8 @@ jobs:
         echo "TRITON_URI=10.1.0.113:5504" >> $GITHUB_ENV
         echo "REDIS_HOST=10.1.0.113" >> $GITHUB_ENV
         echo "REDIS_PORT=6379" >> $GITHUB_ENV
+        echo "KEYCLOAK_OIDC_CONFIG_URL=http://auth.openfoodfacts.org/realms/openfoodfacts/.well-known/openid-configuration" >> $GITHUB_ENV
+        echo "KEYCLOAK_AUDIENCE=account" >> $GITHUB_ENV
     - name: Wait for docker image container build workflow
       uses: tomchv/wait-my-workflow@v1.1.0
       id: wait-build
@@ -165,6 +169,8 @@ jobs:
           echo "REDIS_PRODUCT_UPDATES_STREAM_NAME=product_updates" >> .env
           echo "ENABLE_REDIS_UPDATES=True" >> .env
           echo "Q2_WORKERS=8" >> .env
+          echo "KEYCLOAK_OIDC_CONFIG_URL=${{ env.KEYCLOAK_OIDC_CONFIG_URL }}" >> .env
+          echo "KEYCLOAK_AUDIENCE=${{ env.KEYCLOAK_AUDIENCE }}" >> .env
 
     - name: Create Docker volumes
       uses: appleboy/ssh-action@master
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,6 +37,8 @@ x-api-common: &api-common
     - REDIS_PORT
     - REDIS_PRODUCT_UPDATES_STREAM_NAME
     - REDIS_LATEST_ID_KEY
+    - KEYCLOAK_OIDC_CONFIG_URL
+    - KEYCLOAK_AUDIENCE
   networks:
     - default
     - common_net


open_prices/api/auth/views.py
```diff
@@ -44,6 +47,35 @@
...
+                    "is_moderator": user.is_moderator,
+                    "access_token": access_token,
+                    "token_type": "bearer",
+                }
+            )
+            if request.GET.get("set_cookie") == "1":
+                # Don't add httponly=True or secure=True as it's still in
+                # development phase, but it should be added once the front-end
+                # is ready
+                response.set_cookie(settings.SESSION_COOKIE_NAME, access_token)
```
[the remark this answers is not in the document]
TTalex: We could indeed use a locally generated token, since this is what we do with the username/password flow.

I liked the idea of storing the access_token somewhere, but it's not necessary, since it's unused after the login step for now. We might need it later to auth queries to Product Opener, but it's out of scope as mentionned.

If this is also a security fix, it's probably a good suggestion.

open_prices/api/auth/serializers.py
```diff
@@ -2,8 +2,24 @@
 
 
 class LoginSerializer(serializers.Serializer):
-    username = serializers.CharField()
-    password = serializers.CharField()
+    username = serializers.CharField(required=False)
+    password = serializers.CharField(required=False, write_only=True)
+    access_token = serializers.CharField(required=False, write_only=True)
+
+    def validate(self, data):
```
raphodn, line 9: not a blocker, but not a fan of having this (or any) custom logic in the serializer..
sure the `validate` is part of the Django magic.
but here we have a custom login, i would rather move this in a specific method in `common/authentication.py` with comments + tests + error managed by the view. 
but maybe it's not Django-ic ^^
anyway can be done in another PR

open_prices/api/auth/views.py
```diff
@@ -44,6 +47,35 @@ def post(self, request: Request) -> Response:
         e.g.: "Authorization: bearer token"
         - use the **session** cookie, e.g.: "Cookie: session=token"
         """
+        serializer = self.serializer_class(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        if request.POST.get("access_token"):
+            access_token = request.POST.get("access_token")
```
raphodn: what's the expiration time of the keycloak access_token ?

open_prices/api/auth/views.py
```diff
@@ -44,6 +47,36 @@ def post(self, request: Request) -> Response:
         e.g.: "Authorization: bearer token"
         - use the **session** cookie, e.g.: "Cookie: session=token"
         """
+        serializer = self.serializer_class(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        if request.POST.get("access_token"):
```
raphodn, line 53: todo : room for refactoring (code duplicated with below)

PR #1239, opened 2026-02-27
chore(docs): update INSTALL.md and CONTRIBUTING.md to reflect switch to uv

Emchi24:
This PR updates the [INSTALL.md](https://github.com/openfoodfacts/open-prices/blob/main/INSTALL.md) and [CONTRIBUTING.md](https://github.com/Emchi24/open-prices/blob/main/CONTRIBUTING.md) documentation to align with the recent changes to uv as the package manager by adding the ```uv run``` prefix to the relevant commands.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -25,8 +25,6 @@ cd open-prices
 # Install dependencies (pyproject.toml)
 uv sync
 
-# Note: all future commands should be prefixed with `uv run`
-
 # Apply migrations
 uv run python manage.py migrate
 


INSTALL.md
```diff
@@ -28,15 +28,15 @@ uv sync
 # Note: all future commands should be prefixed with `uv run`
```
TTalex: This line was telling users to use `un run` for all following commands.

I guess it can be removed if we add the `uv run` to the commands :)

Emchi24: Hi, sorry for my late response. You are right, i deleted the line.

PR #1247, opened 2026-03-07
feat: allow use create price-tags without ai

NancyWei123:
### What
- create use_ai to record whether use choose to create price tags without ai
### Fixes bug(s)
- https://github.com/openfoodfacts/open-prices-frontend/issues/1974


Freso (approved):
Seems good to me. 👍

Would be good to have tests and documentation for the parameter as well.

NancyWei123:
Cool! I have added test and doc!
image
image

open_prices/api/proofs/views.py
```diff
@@ -279,8 +279,9 @@ def create(self, request: Request, *args, **kwargs):
 
         user_id = self.request.user.user_id
         price_tag = serializer.save(updated_by=user_id, created_by=user_id)
+        use_ai = request.data.get("use_ai", False)
 
-        if not settings.TESTING:
+        if not settings.TESTING and use_ai:
```
raphodn, line 284: sorry for taking so long to review.
thanks for the PR.
i think we're not ready yet to work on this, what i mean is that we probably need to discuss implementation before merging anything.
my intuition would say that it is a field that should be stored.
similar to `draft` or `ready_for_price_tag_validation`

NancyWei123: Thanks for the feedback. I’ll keep this in mind next time.

PR #1261, opened 2026-03-31
feat: Compare prices against all category children

### LINKED ISSUE #1257: Challenges should include sub-categories for bulk products
### Story

Freso:
### What
- Rough, untested implementation for checking whether a price belongs to not just prices directly in the challenge.categories but also any children of these.

### Fixes bug(s)

Fixes: https://github.com/openfoodfacts/open-prices/issues/1257

### DIFF
--- a/open_prices/challenges/models.py
+++ b/open_prices/challenges/models.py
@@ -10,6 +10,7 @@ from django.utils import timezone
 from open_prices.challenges import constants as challenge_constants
 from open_prices.challenges import validators as challenge_validators
 from open_prices.common import utils
+from open_prices.common.openfoodfacts import get_category_children
 from open_prices.locations.models import Location
 
 
@@ -121,6 +122,16 @@ class Challenge(models.Model):
     @property
     def tag(self):
         return f"{challenge_constants.CHALLENGE_TAG_PREFIX}{self.id}"
+
+    @property
+    def categories_with_children(self) -> list[str]:
+        """A list of a challenge’s categories including the children categories of these."""
+        all_children = self.categories.copy()
+        for category in self.categories:
+            all_children.append(get_category_children(category))
+
+        # Turn into a set before returning to drop duplicates
+        return list(set(all_children))
 
     def location_id_list(self):
         return list(self.locations.values_list("id", flat=True))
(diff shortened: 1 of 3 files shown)


open_prices/common/openfoodfacts.py
```diff
@@ -87,6 +87,30 @@ def normalize_taxonomized_tags(taxonomy_type: str, value_tags: list[str]) -> lis
     return [mapped_tags[k] for k in mapped_tags]
 
 
+def get_category_children(parent: str) -> list[str]:
```
Freso, line 90: I’m not sure how expensive this function is, but given that it’s deterministic, it should be possible to cache it if needed.

(~~I’ll also make a PR upstream against off-python to have this be a `@property` method of `TaxonomyNode`s directly.~~ The logic is mostly copied from `TaxonomyNode.get_parents_hierarchy()`. See also https://github.com/openfoodfacts/openfoodfacts-python/pull/467 )

open_prices/challenges/models.py
```diff
@@ -122,6 +123,16 @@ def status(self) -> str:
     def tag(self):
         return f"{challenge_constants.CHALLENGE_TAG_PREFIX}{self.id}"
 
+    @property
+    def categories_with_children(self) -> list[str]:
```
Freso, line 127: This should also be deterministic if it’d be better to cache here instead of (or in addition to) `common.off.get_category_children()`.

open_prices/prices/models.py
```diff
@@ -119,11 +119,11 @@ def in_challenge(self, challenge: Challenge):
             queryset = queryset.filter(
                 Q(
                     type=price_constants.TYPE_CATEGORY,
-                    category_tag__in=challenge.categories,
+                    category_tag__in=challenge.categories_with_children,
```
raphodn, line 122: my worry is that the `in_challenge` model method is called on each price creation. and so will the categories_with_children

raphodn: also, there is a `in_challenge` queryset, that is used in `set_price_tags` (also `set_proof_tags`) when the weekly task runs to update product changes

Freso: What if `.categories_with_children` (and/or `.get_category_children()`) was cached? I don’t know what the pros/cons would be of caching vs. storing in the database (which, i guess, is just a different kind of caching).

raphodn:
Closing as this has been done in other PRs (#1269, #1270, #1273). 
But heavily inspired from this work :+1:

Freso:
(trying do a thing, sorry for any potential notification spam :eyes:)

PR #1269, opened 2026-04-12
feat(OpenFoodFacts): new taxonomy helper method to get the full list of children from a parent list

### LINKED ISSUE #1257: Challenges should include sub-categories for bulk products

raphodn:
### What

See issue #1257
And with inspiration from a related PR #1261

New taxonomy helper method to extract the full chlidren list from a parent list.

Will be useful for challenges.

### Notes

- extra param `include_parent_list` to be sure to keep the parent as well sometimes
- it currently only takes 1 taxonomy at a time. so if the parent_list contain ids from different taxonomies, it won't be able to fetch all of their children

### Test

```
import time
from open_prices.common import openfoodfacts as common_openfoodfacts

start = time.time()
res = common_openfoodfacts.get_taxonomy_children_tags_from_parent_list("category", ["en:snacks", "en:office-supplies"], include_parent_list=True)
print(len(res))
end = time.time()
print(f"Execution time: {end - start}")

845
Execution time: 0.006926298141479492
```

### DIFF
--- a/open_prices/common/openfoodfacts.py
+++ b/open_prices/common/openfoodfacts.py
@@ -92,6 +92,8 @@ def get_taxonomy_children_tags_from_parent_list(
     :param parent_list: A list of parent tags (e.g., ["en:beverages"]).
     :param include_parent_list: Whether to include the parent tags in the result.
     :return: A list of all children tags (e.g., ["en:beverages", "en:sodas", "en:juices"]).
+
+    TODO: manage parent_list categories coming from different taxonomies (food & non-food)
     """
     taxonomy = _cached_get_taxonomy(taxonomy_type)
     children_tags = [] if not include_parent_list else list(parent_list)
@@ -100,9 +102,9 @@ def get_taxonomy_children_tags_from_parent_list(
     for parent in parent_list:
         parent_node = taxonomy[parent]
 
-        if not parent_node:
-            continue
-        if not parent_node.children:
+        # the parent node might not be in this taxonomy (e.g. non-food)
+        # the parent node might not have children
+        if not (parent_node and parent_node.children):
             continue
 
         for child_node in parent_node.children:
--- a/open_prices/common/openfoodfacts.py
+++ b/open_prices/common/openfoodfacts.py
@@ -94,7 +94,7 @@ def get_taxonomy_children_tags_from_parent_list(
     :return: A list of all children tags (e.g., ["en:beverages", "en:sodas", "en:juices"]).
     """
     taxonomy = _cached_get_taxonomy(taxonomy_type)
-    children_tags = [] if not include_parent_list else parent_list
+    children_tags = [] if not include_parent_list else list(parent_list)
     seen = set()
 
     for parent in parent_list:
--- a/open_prices/common/openfoodfacts.py
+++ b/open_prices/common/openfoodfacts.py
@@ -95,7 +95,7 @@ def get_taxonomy_children_tags_from_parent_list(
     """
     taxonomy = _cached_get_taxonomy(taxonomy_type)
     children_tags = [] if not include_parent_list else list(parent_list)
-    seen = set()
+    seen = set(children_tags)
 
     for parent in parent_list:
         parent_node = taxonomy[parent]
--- a/open_prices/common/openfoodfacts.py
+++ b/open_prices/common/openfoodfacts.py
@@ -53,15 +53,11 @@ _cached_get_taxonomy = functools.lru_cache()(get_taxonomy)
 def normalize_taxonomized_tags(taxonomy_type: str, value_tags: list[str]) -> list[str]:
     """Normalizes a list of tags based on the taxonomy type.
 
-    :param taxonomy_type: The type of taxonomy ('category', 'label', or
-        'origin').
-    :param value_tags: A list of tag values to normalize (e.g.,
-        ["fr: Boissons"]).
-    :raises RuntimeError: If the taxonomy type is not one of 'category',
-        'label', or 'origin'
+    :param taxonomy_type: The type of taxonomy ('category', 'label', or 'origin').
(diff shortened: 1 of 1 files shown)


open_prices/common/openfoodfacts.py
```diff
@@ -87,6 +83,43 @@ def normalize_taxonomized_tags(taxonomy_type: str, value_tags: list[str]) -> lis
...
+    children_tags = [] if not include_parent_list else parent_list
+    seen = set()
+
+    for parent in parent_list:
+        parent_node = taxonomy[parent]
+
+        if not parent_node:
+            continue
+        if not parent_node.children:
+            continue
```
Freso:
```suggestion
        if not (parent_node and parent_node.children):
            continue
```
Python evaluates `and`/`or` statements sequentially, so if `parent_node` evaluates to `False` it won’t look at `.children` and shouldn’t raise any errors.

Freso (approved):
LGTM! (But then, apart from some minor changes, this is more or less identical to https://github.com/openfoodfacts/open-prices/pull/1261 so would probably be good to have someone else’s eyes on it too. :))

open_prices/common/openfoodfacts.py
```diff
@@ -87,6 +83,43 @@ def normalize_taxonomized_tags(taxonomy_type: str, value_tags: list[str]) -> lis
...
+    :return: A list of all children tags (e.g., ["en:beverages", "en:sodas", "en:juices"]).
+    """
+    taxonomy = _cached_get_taxonomy(taxonomy_type)
+    children_tags = [] if not include_parent_list else parent_list
+    seen = set()
+
+    for parent in parent_list:
+        parent_node = taxonomy[parent]
+
+        if not parent_node:
```
Freso: In what case would `parent_node = taxonomy[parent]` set `parent_node` to a falsy value and not throw an error/exception?

open_prices/common/openfoodfacts.py
```diff
@@ -87,6 +83,43 @@ def normalize_taxonomized_tags(taxonomy_type: str, value_tags: list[str]) -> lis
...
+            continue
+        if not parent_node.children:
+            continue
+
+        for child_node in parent_node.children:
+            if child_node.id not in seen:
+                children_tags.append(child_node.id)
+                seen.add(child_node.id)
```
Freso, line 113: Since parents are not in `seen`, does this potentially duplicate parent entries if `include_parent_list=True` and some `parent_list` items are children of other items in the list?

open_prices/common/openfoodfacts.py
```diff
@@ -87,6 +83,43 @@ def normalize_taxonomized_tags(taxonomy_type: str, value_tags: list[str]) -> lis
...
+    """Gets a list of all children for a given taxonomy type and parent list.
+
+    :param taxonomy_type: The type of taxonomy ('category', 'label', 'origin'...).
+    :param parent_list: A list of parent tags (e.g., ["en:beverages"]).
+    :param include_parent_list: Whether to include the parent tags in the result.
+    :return: A list of all children tags (e.g., ["en:beverages", "en:sodas", "en:juices"]).
+    """
+    taxonomy = _cached_get_taxonomy(taxonomy_type)
+    children_tags = [] if not include_parent_list else parent_list
+    seen = set()
```
Freso:
```suggestion
    # Populate `seen` so parents won’t get added multiple times
    seen = set(children_tags)
```

`seen` doesn’t control whether we traverse the tree, only whether we add children to `children_tags` or not, so it should be safe to prepopulate it with any `children_tags` we start off with.

raphodn: good catch !

open_prices/common/openfoodfacts.py
raphodn: yeah you're right the v1 is probably too verbose. 
i had an issue with non-food categories.

open_prices/common/openfoodfacts.py
raphodn: done here: https://github.com/openfoodfacts/open-prices/pull/1269/commits/1a9558555cbb9bdfa790942b7308f2b263a88463

open_prices/common/openfoodfacts.py
raphodn: it seems like it fails silently. so if the parent is not found, taxonomy[parent] returns None

open_prices/common/openfoodfacts.py
raphodn: fixed in https://github.com/openfoodfacts/open-prices/pull/1269/commits/d85a57c6e6077065fbb6068dd8c3d82188c92ade :+1:

open_prices/common/openfoodfacts.py
```diff
@@ -87,6 +83,45 @@ def normalize_taxonomized_tags(taxonomy_type: str, value_tags: list[str]) -> lis
...
+
+    :param taxonomy_type: The type of taxonomy ('category', 'label', 'origin'...).
+    :param parent_list: A list of parent tags (e.g., ["en:beverages"]).
+    :param include_parent_list: Whether to include the parent tags in the result.
+    :return: A list of all children tags (e.g., ["en:beverages", "en:sodas", "en:juices"]).
+
+    TODO: manage parent_list categories coming from different taxonomies (food & non-food)
+    """
+    taxonomy = _cached_get_taxonomy(taxonomy_type)
+    children_tags = [] if not include_parent_list else list(parent_list)
```
Freso, line 99:
```suggestion
    children_tags = [] if not include_parent_list else parent_list.copy()
```
should be cheaper (and also more self-documenting what’s going on even if one doesn’t know Python) since it can potentially be smarter about how to copy it, depending on Python implementation.

raphodn: Will do the change in a future pr 👍

raphodn: done in #1274

PR #1273, opened 2026-04-12
refactor(Challenges): match CATEGORY prices on categories_full (instead of categories)

### LINKED ISSUE #1257: Challenges should include sub-categories for bulk products

raphodn:
### What

Following #1270 (new Categories.categories_full field).

We use this new field for price matching.
- only TYPE_CATEGORY prices (we consider that TYPE_PRODUCT product data has always the full category hierarchy)
- requires to update both the in_challenge queryset & property

### DIFF
--- a/open_prices/prices/models.py
+++ b/open_prices/prices/models.py
@@ -111,6 +111,12 @@ class PriceQuerySet(models.QuerySet):
         return self.filter(tags__contains=[tag])
 
     def in_challenge(self, challenge: Challenge):
+        """
+        Return prices that are in the given challenge, based on:
+        - the price creation date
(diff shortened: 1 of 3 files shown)


open_prices/prices/models.py
```diff
@@ -377,38 +383,36 @@ def update_tags(self):
             self._change_reason = "Price.update_tags() method"
             self.save(update_fields=["tags"])
 
-    def has_category_tag(self, category_tag_list: list):
```
raphodn, line 380: removed has_category_tag
- was not used elsewhere except in_challenge
- clarified the in_challenge property instead

PR #1275, opened 2026-04-13
refactor(Challenges): improve daily update task to update categories_full & ignore past (completed) challenges

raphodn:
### What

We already have a nightly task that updates EVERY challenge.

Following #1270 we now have a `Challenge.categories_full` field, that we would like to update regularly as the taxonomy evolves.

This PR changes the nightly task
- to include the `calculate_categories_full` method
- to ignore COMPLETED challenges after a delay of 7 days (freeze stats)

### DIFF
--- a/open_prices/challenges/models.py
+++ b/open_prices/challenges/models.py
@@ -359,4 +359,4 @@ def challenge_post_create_init_categories_full_and_stats(
 ):
     if created:
         instance.calculate_categories_full()
-        instance.calculate_stats()
+        instance.calculate_stats()  # init
--- a/open_prices/challenges/models.py
+++ b/open_prices/challenges/models.py
@@ -56,12 +56,15 @@ class ChallengeQuerySet(models.QuerySet):
 
     def to_update_in_daily_task(self):
         """
-        Goal: freeze completed challenges (categories, tags, stats) after a delay.
+        Goal: Choose which challenges stats we want to update.
+        - Data updated? categories_full, price/proof tags, stats
+        - We stop updating ("freeze") completed challenges after a delay
+
         Usage: see open_prices/common/tasks.py:challenge_tasks
         """
         CHALLENGE_COMPLETED_FREEZE_DELAY_IN_DAYS = 7
         return self.with_status().filter(
-            ~Q(status_annotated=challenge_constants.CHALLENGE_STATUS_COMPLETED)
+            Q(end_date=None)
             | Q(
                 end_date__gte=timezone.now().date()
                 - timedelta(days=CHALLENGE_COMPLETED_FREEZE_DELAY_IN_DAYS)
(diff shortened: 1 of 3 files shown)


open_prices/challenges/models.py
```diff
@@ -54,6 +54,20 @@ def is_ongoing(self):
             status_annotated=challenge_constants.CHALLENGE_STATUS_ONGOING
         )
 
+    def to_update_in_daily_task(self):
+        """
+        Goal: freeze completed challenges (categories, tags, stats) after a delay.
+        Usage: see open_prices/common/tasks.py:challenge_tasks
+        """
+        CHALLENGE_COMPLETED_FREEZE_DELAY_IN_DAYS = 7
```
raphodn, line 65: what should the delay be ?

open_prices/challenges/models.py
```diff
@@ -54,6 +54,20 @@ def is_ongoing(self):
...
+        return self.with_status().filter(
```
raphodn, line 66: this could be simplified i guess, and just filter on the end_date ?

open_prices/challenges/models.py
TTalex: 7 days is as good a guess as any. We'll see in practice.
We could also consider manually triggering the update for a specific challenge later on, bypassing the freeze, if we really notice something bad.

open_prices/challenges/models.py
TTalex: Maybe having CHALLENGE_STATUS_COMPLETED will act as a safeguard for old draft challenges that were never run ?

open_prices/challenges/models.py
TTalex: Should that be a settings.py param ?

TTalex (approved):
Nice !

open_prices/challenges/models.py
Freso: I think making it a setting would make sense, but that can also go in its own, later, PR.

open_prices/challenges/models.py
Freso: I don’t think that’s a safeguard that needs to be there. If it’s an old draft challenge that’s `ended` more than `CHALLENGE_COMPLETED_FREEZE_DELAY_IN_DAYS` days ago, we also don’t want to include it with `to_update_in_daily_task()` regardless, right?

Freso (approved):
LGTM! Two potential improvements in comments, but nothing I would say that blocks this PR outright.

open_prices/challenges/models.py
raphodn: in any case we can connect to the server and launch stuff manually in a shell

open_prices/challenges/models.py
raphodn: ok i simplified to filter only on the end_date : https://github.com/openfoodfacts/open-prices/pull/1275/commits/f40404f2d91f2ed6b83ee3e804111e89e928beea
(and keep the case where end_date is empty, e.g. draft challenges, where we might want to see the categories_full evolve on changes (challenge.categories or taxonomy))

PR #1289, opened 2026-05-01
feat(Stats): new product_created fields (total & per flavor) (filter on 'open-prices' creator)

### LINKED ISSUE #1280: Stats: add the number of products created by 'open-prices' user
### Story

Following #1084
Add a new stat to count the number of products added (OxF ? per flavor ?)

raphodn:
### What

Following #1288 (`Product.creator field)

New `Stats.product_created` fields, for the user `open-prices`

```
product_created_count
product_created_source_off_count
product_created_source_obf_count
product_created_source_opff_count
product_created_source_opf_count
```


open_prices/stats/models.py
```diff
@@ -123,6 +125,7 @@ class TotalStats(SingletonModel):
     user_count = models.PositiveIntegerField(default=0)
     user_with_price_count = models.PositiveIntegerField(default=0)
     challenge_count = models.PositiveIntegerField(default=0)
+    off_contributions_stats = models.JSONField(default=dict)
```
raphodn: not sure if having a JSONField is necessarily a good idea

Positive
- more flexible, allows to add metadata such as the user_id
- if we also add a set of "updated" stats it'll be easy to add
- if we change to giving the user authorship, we'll just delete the user_id key

Negative
- well we could simply add the IntegerFields directly and it wouldn't change much...
- API discoverability, the JSON looks empty

raphodn: @TTalex what do you think ? :sweat_smile:

TTalex: JSONField looks fine to me :
- We know that authorship might change at some point
- Maybe we'll also track edits in a future version.
- I like the idea of bundling these stats that are not strictly related to the Open Prices core features (opens up a future `osm_contributions_stats` ?)
- I don't think API discoverability matters, this is not an endpoint I expect many apps to use, other than a different frontend.

open_prices/stats/models.py
```diff
@@ -260,3 +264,20 @@ def update_challenge_stats(self):
 
         self.challenge_count = Challenge.objects.published().count()
         self.save(update_fields=self.CHALLENGE_COUNT_FIELDS + ["updated"])
+
+    def update_other_stats(self):
```
TTalex: Would it be cleaner to keep this function as `update_off_contributions_stats` instead of the generic "other_stats" ?

If we add other random stats in the future, I'm guessing it would be easy to create a dedicated method, add it clearly to the tasks file and unit tests will be easier to write independently of the current off_contributions_stats ones.

TTalex (approved):

open_prices/stats/models.py
raphodn: i talked a bit with copilot this afternoon and i kinda feel adding IntegerField is better in the end ^^

open_prices/stats/models.py
raphodn: good point :+1:

open_prices/stats/models.py
raphodn: did the switch in https://github.com/openfoodfacts/open-prices/pull/1289/commits/0a076c85de76b8432ca0e6bb9334a737299f5c7f

open_prices/stats/models.py
raphodn: done in https://github.com/openfoodfacts/open-prices/pull/1289/commits/0a076c85de76b8432ca0e6bb9334a737299f5c7f

PR #1292, opened 2026-05-05
feat(Proofs): API: new endpoint to upload (and edit) "draft" proofs

### LINKED ISSUE #957: Allow to anonymize receipts
Sometimes, receipts contain personal data, such as:

- the name of the buyer
- the fidelity ID
- the hour of purchase (this can be considered a PII as we can know who uploaded the proof using for example videos from the shop)

It would be very valuable to provide a way to anonymize receipts automatically (using visual LLMs for example).
The user would upload the proof and the ML model would automatically remove using black squares the sensitive information. The user would also be able to correct the LLM anonymization manually if needed.

### LINKED ISSUE #958: Store uploaded images temporarily on server
Currently, when a user uploads a proof, it's only sent to the server when the user clicks on the "Upload" button, along with all metadata provided by the user. The proof is immediately available in the `GET /proofs` route.

This is not ideal, for several reasons:

- it makes it harder to anonymize data without sending the image twice, once for analysis and once for upload, or without making publicly available the proof before anonymization (see #957)
- it prevents us from leveraging ML models running server-side to help pre-fill the data (ex: location, date, suggesting a different proof type if the user made a mistake)

## Suggestion

We could send the image as soon as the user selects an image to be uploaded. This image would still be stored in the `proof` table, but under a special status: we would add a `draft` field that's equal to `true`.

Draft proofs wouldn't be visible by anyone: they exist just for the time for the user to finish uploading the proof.
After a reasonable time without being completed (ex: 1h), they are deleted automatically from DB.

At first, we only upload the proof, without associated metadata. We can immediately run the ML models on this proof, which help us in the front-end to leverage these predictions.

For example, for proof anonymization, we can detect the PII in the proof, and allow users to visualize (and modify) the redacted parts immediately after upload.

Once the proof is submitted with associated metadata, we remove the draft flag (`draft=false`).
The proof becomes immediately visible by anyone.

I'm interested in having feedback on this :)
@raphodn @TTalex

raphael0202:
Here, we add a new `proof.draft` flag, that indicates that the proof is a "draft" (unfinished).

Draft proofs are not visible to any user but the user who uploaded it.

We only require for now the proof type when creating the draft proof, all other fields are optional. The advantage of draft proofs is to allow submitting immediately the image, as soon as the user select the proof to upload. Another advantage is that we run ML models immediately after upload. These model could be leveraged to help the user complete some proof-related information:

- date, number of items, total price, location (for receipts)
- currency

This also unlock the possibility to anonymize receipts, by first uploading the photo in draft mode (not available to other users), and then adding black rectangles to areas to anonymize.

Additional information (including currency, date and optional location) are send using the PATCH /proofs/finalize endpoint, the draft flag is removed during this step.

Draft proofs are deleted 1h after their creation date.

Related issue: https://github.com/openfoodfacts/open-prices/issues/958

Unit & integration tests were not added yet, I wanted to validate the features first.

### DIFF
--- a/open_prices/api/proofs/views.py
+++ b/open_prices/api/proofs/views.py
@@ -1,12 +1,16 @@
 import PIL.Image
 from django.conf import settings
+from django.db.models import Q
 from django_filters.rest_framework import DjangoFilterBackend
 from django_q.tasks import async_task
 from drf_spectacular.utils import extend_schema
 from rest_framework import filters, mixins, status, viewsets
 from rest_framework.decorators import action
 from rest_framework.parsers import MultiPartParser
-from rest_framework.permissions import IsAuthenticatedOrReadOnly
+from rest_framework.permissions import (
+    IsAuthenticated,
+    IsAuthenticatedOrReadOnly,
+)
 from rest_framework.request import Request
 from rest_framework.response import Response
 
@@ -21,6 +25,7 @@ from open_prices.api.proofs.serializers import (
     PriceTagFullSerializer,
     PriceTagUpdateSerializer,
     ProofCreateSerializer,
+    ProofDraftUploadSerializer,
     ProofFullSerializer,
     ProofHalfFullSerializer,
     ProofHistorySerializer,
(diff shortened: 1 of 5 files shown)


open_prices/api/proofs/views.py
```diff
@@ -63,9 +65,6 @@ class ProofViewSet(
     ordering = ["id"]
 
     def get_authenticators(self):
-        if self.request and self.request.method in ["GET"]:
```
raphael0202, line 66: I'm not sure to understand why we used to fallback to the default authenticator (none) here. We now need to know if the user is authenticated to list proofs.

raphodn: currently, anonymous users can fetch all proofs (and all prices). changing this line would break that I believe

open_prices/api/proofs/views.py
```diff
@@ -101,6 +106,21 @@ def destroy(self, request: Request, *args, **kwargs) -> Response:
         permission_classes=[],  # allow anonymous upload
     )
     def upload(self, request: Request) -> Response:
+        return self.base_upload(request, draft=False)
+
+    @extend_schema(request=ProofDraftUploadSerializer, responses=ProofFullSerializer)
+    @action(
+        detail=False,
+        methods=["POST"],
+        url_path="draft",
```
raphodn:
```suggestion
        url_path="upload/draft",
```
more understandable ?

open_prices/api/proofs/views.py
```diff
@@ -246,12 +327,21 @@ def get_authenticators(self):
         return [CustomAuthentication()]
 
     def get_queryset(self):
+        # Filter out draft proofs for non-owners
```
raphodn: another option is to not run PriceTag stuff on draft proofs.
that will clarify the workflow

draft --> AI x proof models --> finalize --> AI x price models

open_prices/api/proofs/views.py
raphodn: actually we could keep `/draft` and have everything under
- with a dedicated queryset that includes draft=True
- users need to be owners
- RESTful : GET, POST (/upload ?), PATCH, DELETE

open_prices/api/proofs/views.py
```diff
@@ -74,6 +73,12 @@ def get_queryset(self):
             queryset = queryset.select_related("location")
             if self.action == "retrieve":
                 queryset = queryset.prefetch_related("predictions")
+        # Filter out draft proofs for non-owners
+        if self.request.user.is_authenticated:
+            user_id = self.request.user.user_id
+            queryset = queryset.filter(Q(draft=False) | Q(owner=user_id))
```
raphodn: - if we have a `/draft` dedicated endpoint this would be moved there
- and it could be simplified + moved this logic to the model
have the default `Proof.objects.all()` that hides the draft=True, and a new queryset `Proof.objects_with_draft.all()` only in the `/draft` endpoint

open_prices/api/proofs/views.py
raphael0202: From my understanding [of the docs](https://www.django-rest-framework.org/api-guide/authentication/), we should authenticate all requests and use the permission system to allow (or not) requests.
But as we're going to implement a different class for draft proofs, this is not the subject and I'm reverting the changes.

open_prices/api/proofs/views.py
raphael0202: Good idea, I'm creating a new `DraftViewSet` ViewSet.

open_prices/api/proofs/views.py
raphael0202: As discussed, I'm keeping the `/upload` suffix to keep the same convention as for proofs.

open_prices/api/proofs/views.py
raphael0202: I think the quicker we run the ML models (including price tag detection and extraction), the better the user experience, so I would rather keep it this way!

open_prices/api/proofs/views.py
```diff
@@ -69,7 +167,8 @@ def get_authenticators(self):
         return [CustomAuthentication()]
 
     def get_queryset(self):
-        queryset = self.queryset
+        # Filter out draft proofs, as we use a dedicated draft endpoint to handle them
+        queryset = self.queryset.filter(draft=False)
```
raphodn: it could be set line 156 instead 😉

open_prices/api/proofs/views.py
```diff
@@ -35,13 +40,106 @@
...
+    OnlyObjectOwnerIsAllowed,
+    OnlyObjectOwnerOrModeratorIsAllowed,
+)
 from open_prices.proofs import constants as proof_constants
 from open_prices.proofs.ml.price_tags import extract_from_price_tag
 from open_prices.proofs.models import PriceTag, Proof, ReceiptItem
 from open_prices.proofs.utils import compute_file_md5, store_file
 
 
+def base_upload(request: Request, draft: bool = False) -> Response:
```
raphodn, line 52: 👍

open_prices/api/proofs/views.py
```diff
@@ -246,12 +346,21 @@ def get_authenticators(self):
         return [CustomAuthentication()]
 
     def get_queryset(self):
+        # Filter out draft proofs for non-owners
+        if self.request.user.is_authenticated:
+            user_id = self.request.user.user_id
+            base_queryset = self.queryset.filter(
```
raphodn:
```suggestion
            queryset = self.queryset.filter(
```
to homogenize with the namings used elsewhere

open_prices/common/permission.py
```diff
@@ -15,12 +15,10 @@ def request_user_is_moderator(request) -> bool:
...
 class OnlyObjectOwnerIsAllowed(BasePermission):
     """
-    Only give access to object owners.
+    Only give access to object owners, for all methods (GET, POST, PUT, PATCH, DELETE).
     """
 
-    def has_object_permission(self, request, view, obj):
-        if request.method in SAFE_METHODS:
-            return True
+    def has_object_permission(self, request, view, obj) -> bool:
```
raphodn: that seems like a change that merits a dedicated PR..

open_prices/proofs/models.py
```diff
@@ -206,6 +206,8 @@ class Proof(models.Model):
     owner_consumption = models.BooleanField(blank=True, null=True)
     owner_comment = models.TextField(blank=True, null=True)
 
+    draft = models.BooleanField(default=False)
```
raphodn, line 239: reading the PR, users can add prices on draft proofs ?? i wouldn't allow it

raphodn: and we could have filtered on just receipts to start with.. is there a point of having draft price tags ?

raphodn: following discussion
- adding prices on draft proofs is not wanted
- any kind of proof can be set to draft, it'll be deleted in 1h in any case
- extra work can be done in another PR

open_prices/common/permission.py
raphodn: or maybe i'm not understanding the change correctly ?

raphodn (approved):

open_prices/proofs/models.py
raphodn: ok I opened #1326 to manage the model changes. this PR stays to manage the API changes.

raphodn: I'm going to merge + rebase

raphodn: before: 9 files changed / +465-101
after: 5 files changed / +422-98

open_prices/api/proofs/views.py
raphodn: I reverted the changes here.
sure we can run the PriceTag stuff on proof upload.
but we should wait for the proof to be non-draft before allowing users to view and add prices.

open_prices/common/permission.py
raphodn: did some digging, opened this PR to clarify stuff: #1327

raphodn: ok cleaned up :)

open_prices/api/proofs/views.py
raphodn: not needed anymore following #1326

PR #1318, opened 2026-05-18
fix: improve error handling when fetching OFF product details

raphael0202:
- update `get_product_dict` docstring to document explicitely the func behaviour: return None when API call fails and an empty dict when the product was not found.
- log exceptions in `get_product_dict` when fetching product data from OFF API
- In `build_product_dict`, only catch exceptions when calling OFF API: previously, exceptions in `build_product_dict` were silently catched and ignored.


TTalex (approved):
Looks good to me

PR #1284, opened 2026-04-24
feat: integrate price tag classification model

raphael0202:
This PR integrates the price tag classification model.
Each detected price tag is classified into one of the following class: `high-quality`, `medium-quality` or `invalid`.
Data extraction (price, barcode,...) is only run on price tags that have a type different from `invalid`.

As a new value for price_tag.type DB field was added, `make makemigrations` was run.

Associated front-end PR: https://github.com/openfoodfacts/open-prices-frontend/pull/2174

PR #1328, opened 2026-06-01
fix(Price Tags): fix properties and tags computation when multiple prediction

### LINKED ISSUE #1304: New ML model to classify price tag data
### Story

Currently, we send all detected price tags to Gemini, even though some might be unreadable (truncated, blurry...)

The idea here is to run a model beforehand to classify the image.
And then decide if it has the quality threshold to be sent to the following "extraction" step

### Context

The extraction step is done by Gemini.
There is ongoing work to build an alternate model (#931) + integrate it in Open Prices (TBD)

TTalex:
Following #1284

### What
- I've noticed the frontend filter tag_prediction_product_exists was no longer working (showing no price tags)
- My guess is that the tag computation runs on the wrong prediction (clf instead of extraction)
- Indeed, the code is always looking for the `first` prediction, which now is the classification one
- I've also noticed that some of the PriceTag properties share the same broken predictions.first logic
- This PR properly loads the correct extraction prediction, if available, instead of the first one

Should fix the frontend display issue https://github.com/openfoodfacts/open-prices-frontend/issues/2216

### DIFF
--- a/open_prices/proofs/models.py
+++ b/open_prices/proofs/models.py
@@ -664,9 +664,7 @@ class PriceTag(models.Model):
         return False
 
     def get_prediction_from_type(self, type: str):
-        if self.predictions.exists():
-            return self.predictions.filter(type=type).first()
-        return None
+        return self.predictions.filter(type=type).first()
 
     def update_tags(self):
         changes = False
--- a/open_prices/proofs/models.py
+++ b/open_prices/proofs/models.py
@@ -677,7 +677,9 @@ class PriceTag(models.Model):
             price_tag_prediction_has_predicted_product_exists,
         )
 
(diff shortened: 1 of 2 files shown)


open_prices/proofs/models.py
```diff
@@ -663,6 +663,11 @@ def set_tag(self, tag: str, save: bool = True):
             return True
         return False
 
+    def get_prediction_from_type(self, type: str):
```
raphodn, line 666: i tried to see how it could be optimized (avoiding N+1 queries), but didn't find easily

raphodn: i simplified to 1 line https://github.com/openfoodfacts/open-prices/pull/1328/commits/fac82f1e23dad6da0046f46a658a104dd4322eb3

raphodn (approved):

PR #1337, opened 2026-06-10
feat: add API rate limiting

### LINKED ISSUE #1301: API Rate Limiting (reduce scraping)
### Story

There's currently no rate limiting set up on the API...
And we're not monitoring usage, but there is probably some people fetching many prices, when they could be getting the export (parquet) data instead.

### How

Look into what has been done on the OFF side.
And Django tooling

saivats:
## What
Fixes #1301 — adds DRF throttling to the API.

### config/settings.py
Added throttle configuration to the REST_FRAMEWORK dict:
- Anonymous users: 60 requests/minute (per IP)
- Authenticated users: 600 requests/minute

### open_prices/api/tests.py
Added RateLimitConfigTests (5 tests, all passing) using SimpleTestCase
to avoid requiring a database connection.

### pagination.py
No change needed — max_page_size = 100 was already present.

## Notes
- Throttle rates (60/min anon, 600/min user) are starting points 
  based on OFF server's approach (anon limited, authenticated users 
  get more headroom). raphodn should tune based on actual traffic data.
- Used DRF's built-in AnonRateThrottle + UserRateThrottle — no new 
  dependencies required.
- Tests run without Docker/Postgres using SimpleTestCase.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -202,6 +202,14 @@ REST_FRAMEWORK = {
     "COERCE_DECIMAL_TO_STRING": False,  # return floats instead
     "DEFAULT_AUTHENTICATION_CLASSES": [],
     "DEFAULT_PERMISSION_CLASSES": [],
+    "DEFAULT_THROTTLE_CLASSES": [
+        "rest_framework.throttling.AnonRateThrottle",
+        "rest_framework.throttling.UserRateThrottle",
+    ],
+    "DEFAULT_THROTTLE_RATES": {
+        "anon": "60/minute",
+        "user": "600/minute",
+    },
 }
 
 SPECTACULAR_SETTINGS = {
--- a/open_prices/api/tests.py
+++ b/open_prices/api/tests.py
@@ -1,6 +1,6 @@
 import base64
 
-from django.test import TestCase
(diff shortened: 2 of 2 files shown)


saivats:
@raphodn  — implemented using DRF's built-in throttling. 
Two changes: throttle config in settings.py + 5 config tests.
Pagination was already capped at max_page_size=100 so no change needed there.
Rates (60/min anon, 600/min user) are starting points — please adjust 
based on actual production traffic.

raphael0202:
Sorry for the late response!
I think it's better to implement rate-limiting on nginx reverse proxy directly, as we do in [openfoodfacts](https://github.com/openfoodfacts/openfoodfacts-infrastructure/blob/develop/confs/scaleway-proxy/nginx/sites-enabled/openfoodfacts.org?plain=1).

saivats:
okay got it I didn't realize the convention was nginx-level for this.
Should I close this PR, or is there value in keeping app-level throttling too
as a secondary safeguard? If it's not needed, happy to close and free up the
issue for whoever picks up the nginx config.

raphael0202:
It's better to remove completely the app-level throttling. You can create a new PR!

PR #1346, opened 2026-06-18
feat(Proofs): cascade delete linked prices when proof is deleted

### LINKED ISSUE #1258: Moderation: Deleting a proof should delete all linked prices for moderators
### Problem
Deleting a proof should delete all linked prices for moderators.
Otherwise, proofs with many prices are very tedious to handle (we have one with hundreds of prices in the backlog)

saivats:
Fixes #1258

## What

When a moderator (or owner) deletes a proof, all linked prices are now
automatically deleted first. Previously the API returned 403 Forbidden
if any prices were still attached, making moderation of proofs with
hundreds of prices very tedious.

## Approach

Modified `destroy()` in `ProofViewSet` to call `proof.prices.all().delete()`
before deleting the proof itself.

Chose view-level deletion over changing `on_delete=CASCADE` on the FK
because the `Price.proof` FK is intentionally nullable (prices can exist
without a proof during creation), and CASCADE at the DB level would affect
admin/shell deletions too broadly.

The existing `price_post_delete_decrement_counts` signal fires correctly
for each deleted price, so User, Product, and Location counters are
decremented automatically with no extra logic needed.

## Changes

- `open_prices/api/proofs/views.py` — replace 403 block with
  `proof.prices.all().delete()` before `super().destroy()`
- `open_prices/api/proofs/tests.py` — rename existing test to reflect
  new cascade behavior, remove manual price cleanup in 3 tests (now
  handled by cascade), add 2 new tests:
  - `test_proof_delete_cascades_multiple_prices`
  - `test_proof_delete_moderator_cascades_prices`

## Tests

399 tests pass. ruff clean.

### DIFF
--- a/open_prices/api/pagination.py
+++ b/open_prices/api/pagination.py
@@ -34,18 +34,22 @@ class CustomPagination(PageNumberPagination):
                 "items": schema,
                 "page": {
                     "type": "integer",
+                    "description": "Current page number",
                     "example": 1,
                 },
                 "pages": {
                     "type": "integer",
+                    "description": "Total number of pages",
                     "example": 16,
                 },
                 "size": {
                     "type": "integer",
+                    "description": "Number of items per page",
                     "example": 100,
                 },
                 "total": {
                     "type": "integer",
+                    "description": "Total number of items",
                     "example": 1531,
                 },
             },
--- a/open_prices/api/pagination.py
+++ b/open_prices/api/pagination.py
@@ -1,4 +1,3 @@
-from drf_spectacular.extensions import OpenApiPaginationExtension
 from rest_framework.pagination import PageNumberPagination
 from rest_framework.response import Response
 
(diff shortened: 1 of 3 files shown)


saivats:
@raphodn — would love your feedback when you get a chance. Happy to adjust if the cascade-for-both-owners-and-moderators approach needs to be scoped down to moderators only.

open_prices/api/pagination.py
TTalex, line 34: Unrelated change, could be removed from PR

TTalex:
Hey, thanks for the PR.

It feels a bit risky to me to remove the guardrail for every user. In the UI, it's pretty easy to delete a proof, cascading it to prices might be a mistake users shouldn't be able to make.

That being said, having moderator specific behavior on deletion would probably lead to pretty messy code, which I do not really like. Especially since this is a very niche feature (very few proofs are concerned, and moderators tend to have other tools to handle such cases).

So let's put this PR on hold for now while we think of the best approach.

saivats:
Makes sense ,completely get your point also if you land on an approach and you would like my help on it, please let me know. I'll try my best.

TTalex:
Hey, to keep you updated, this issue was talked about during last weekly meeting. Here is what was mentioned :

- Allowing all users to delete on cascade is the right approach, no need for moderation only conditions
- However, this backend change must require explicit warning on the frontend (out of scope of this PR)
- We're still unsure how to handle the case where a proof has prices that were added by another user. In that case, cascading would allow the proof owner to delete prices that they do not own, which is prevented otherwise.

saivats:
Thanks for the update, good to know the direction is clearer now.
Happy to wait until the frontend warning and ownership question are
resolved before making changes. Just let me know when you'd like me
to pick it back up.

PR #1347, opened 2026-06-18
feat(Proofs): Skip proof type prediction when no available Triton server

### LINKED ISSUE #1209: Local setup: allow disabling only partly the AI
### Story

Currently, there's no easy way to disable only a part of the AI (Gemini, Vertex)

TBC

raised by @TTalex

saivats:
Fixes #1209

## What

`ENABLE_ML_PREDICTIONS` was all-or-nothing. If set to `True`, the pipeline
tried to run all three ML steps even when Triton or Gemini weren't configured,
causing the whole thing to crash.

Now each step only runs if its required dependency is configured. Missing
a dependency logs a warning and moves on to the next step.

## How

Changed `TRITON_URI` default from `"localhost:5504"` to `None` in settings.
The old default was causing installs without Triton to always attempt a
connection and fail. Installs that actually have Triton already set this
in their env, so they're unaffected.

In the orchestrator `run_and_save_proof_prediction()`:

- Steps 1 and 2 (proof type classification, price tag detection) are gated
  on `TRITON_URI` being set
- Step 3 (receipt extraction) is gated on `GOOGLE_CREDENTIALS` being set
- Price tag extraction sub-step within step 2 passes `run_extraction=False`
  when Gemini isn't available, so detection still runs without it
- Each step is also wrapped in `try/except` as a runtime safety net — a
  network failure in one step won't take down the others

`ENABLE_ML_PREDICTIONS` stays as the master on/off switch, no new env
variables added.

## Tests

Added `GranularMLPipelineTest` with 5 cases covering: Triton steps skipped
when `TRITON_URI` is None, Gemini step skipped when `GOOGLE_CREDENTIALS` is
None, runtime exception in one step doesn't block others, all steps run when
fully configured, price tag detection still runs without Gemini.

13 ML tests pass (5 new + 8 existing). ruff clean.


saivats:
@raphodn @TTalex — implemented the best-effort approach as discussed. Happy to adjust anything, especially around the TRITON_URI default change if that's a concern.

TTalex (approved):
I ended up rolling back most changes in PR.

A few changes happened since this PR creation, making the problem simplier.

In the end, the only action that was still needed was preventing `run_and_save_proof_type_prediction(image, proof)` from raising an error when the triton_uri wasn't set.

Users can now only have Gemini capabilities enabled, and still use the Proof Price Tag and Receipt Assistants without blocking errors.

Thank you for your contribution :).

PR #1354, opened 2026-06-18
fix(Price Tags): Predictions: run extraction even if they are classified as invalid

### LINKED ISSUE #1304: New ML model to classify price tag data

raphael0202:
The classifier wrongfully classified price tags of "raw products" as invalid, even when they're easy to read. We need to retrain the classifier with more example of these price tags, but in the meantime, let's run the extraction on all price tags.

### DIFF
--- a/open_prices/proofs/ml/price_tags.py
+++ b/open_prices/proofs/ml/price_tags.py
@@ -788,8 +788,9 @@ def create_price_tags_from_proof_prediction(
             # on price tags that are not predicted as "invalid" as there is nothing
             # to extract on these image crops: either the price tag is too blurry or
             # the crop is not actually a price tag.
-            if predicted_price_tag_type != "invalid":
-                to_process.append(price_tag_with_image)
+            # 2026-06-18: temporarily always run price tag extractions, even on price
+            # tags that are predicted as "invalid"
+            to_process.append(price_tag_with_image)
         else:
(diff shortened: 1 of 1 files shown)


TTalex (approved):
Looks good to me

Edit: ah, minus the pre-commit error in hooks.
Maybe we could do something like 
```
if predicted_price_tag_type != "invalid":
    pass
```
so we keep track of what the check was, and still use the variable

open_prices/proofs/ml/price_tags.py
```diff
@@ -779,17 +779,17 @@ def create_price_tags_from_proof_prediction(
             )
             if not classification:
                 continue
-            predicted_price_tag_type = classification.data.get("prediction", [{}])[
-                0
-            ].get("label")
+            classification.data.get("prediction", [{}])[0].get("label")
```
raphodn, line 782: i fixed the linting issue + added an additional comment line

raphodn (approved):

PR #1355, opened 2026-06-18
chore(API): set the max page number queryable to 500

### LINKED ISSUE #1301: API Rate Limiting (reduce scraping)

raphodn:
### What

Improve our CustomPagination
- new `max_page_number` set to 500
  - f the page 501 is asked, it will return a 400 with a message (link to the docs [Guides > Data](https://openfoodfacts.github.io/open-prices/guides/data/)).
- also disable the `page=last` default behavior
  - return a 400 with a message ("Invalid page")

Add tests.

### Why

Reduce scraping.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -220,8 +220,8 @@ SPECTACULAR_SETTINGS = {
     "DESCRIPTION": "Open Prices API allows you to add product prices",
     "CONTACT": {
         "name": "The Open Food Facts team",
-        "url": f"{OPENFOODFACTS_URL}",
-        "email": f"{OPENFOODFACTS_EMAIL}",
+        "url": OPENFOODFACTS_URL,
+        "email": OPENFOODFACTS_EMAIL,
     },
     "LICENSE": {
         "name": " AGPL-3.0",
--- a/config/settings.py
+++ b/config/settings.py
@@ -2,6 +2,17 @@ import os
 import sys
 from pathlib import Path
 
+OPENFOODFACTS_URL = "https://world.openfoodfacts.org"
+OPENFOODFACTS_EMAIL = "contact@openfoodfacts.org"
+OPENPRICES_DOCS_URL = "https://openfoodfacts.github.io/open-prices/"
+OPENPRICES_DOCS_GUIDES_DATA_URL = (
+    "https://openfoodfacts.github.io/open-prices/guides/data/"
+)
+
+
+# Django config
+# ------------------------------------------------------------------------------
+
 # Build paths inside the project like this: BASE_DIR / 'subdir'.
 BASE_DIR = Path(__file__).resolve().parent.parent
 
@@ -209,8 +220,8 @@ SPECTACULAR_SETTINGS = {
     "DESCRIPTION": "Open Prices API allows you to add product prices",
     "CONTACT": {
         "name": "The Open Food Facts team",
-        "url": "https://world.openfoodfacts.org",
-        "email": "contact@openfoodfacts.org",
+        "url": f"{OPENFOODFACTS_URL}",
+        "email": f"{OPENFOODFACTS_EMAIL}",
     },
     "LICENSE": {
         "name": " AGPL-3.0",
--- a/open_prices/api/pagination.py
+++ b/open_prices/api/pagination.py
@@ -1,5 +1,5 @@
 from django.conf import settings
-from rest_framework.exceptions import NotFound
+from rest_framework.exceptions import ParseError
 from rest_framework.pagination import PageNumberPagination
 from rest_framework.response import Response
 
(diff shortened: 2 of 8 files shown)


open_prices/api/pagination.py
```diff
@@ -11,9 +13,34 @@ class CustomPagination(PageNumberPagination):
...
     page_size = 10
-    page_size_query_param = "size"
+    page_size_query_param = "size"  # default is None
     max_page_size = 100
+    ### page number config
+    # page_query_param = "page"  # default
+    last_page_strings = ()  # disable "last" page string
+    max_page_number = 1000  # custom rule, see get_page_number()
+
+    def get_page_number(self, request, paginator):
```
raphodn, line 25: we override [get_page_number](https://github.com/encode/django-rest-framework/blob/cf582fb58e9e5ffcc8ed78a2cb9aaa8f4865666a/rest_framework/pagination.py#L215)

open_prices/api/pagination.py
```diff
@@ -11,9 +13,34 @@ class CustomPagination(PageNumberPagination):
...
+        """
+        page_number = super().get_page_number(request, paginator)
+        try:
+            page_number_int = int(page_number)
+        except (TypeError, ValueError):
+            raise NotFound("Invalid page.") from None
+        if page_number_int < 1:
+            raise NotFound("Invalid page.")
+        if page_number_int > self.max_page_number:
+            raise NotFound(
```
TTalex: Maybe rest_framework.exceptions.bad_request instead of NotFound ? So the HTTP code is 400, since it's a user error, not a missing resource ?

config/settings.py
```diff
@@ -209,8 +220,8 @@
     "DESCRIPTION": "Open Prices API allows you to add product prices",
     "CONTACT": {
         "name": "The Open Food Facts team",
-        "url": "https://world.openfoodfacts.org",
-        "email": "contact@openfoodfacts.org",
+        "url": f"{OPENFOODFACTS_URL}",
```
raphael0202: No need to use f-strings here, referencing the string directly does the job.

open_prices/api/pagination.py
raphodn: indeed 404 seems misleading

open_prices/api/pagination.py
raphael0202: I think we should limit the total number of results, not the page number.
Providing a different `size` (ex: size=100) parameter allows fetching more results that with the default page size.

open_prices/api/pagination.py
raphael0202: +1 for the HTTP 400.

open_prices/api/pagination.py
raphodn: there is already a limit on size = 100
so its 2 different things no ?
here we limit on the page number, so indeed max becomes 100x1000 = 100k
we can reduce the max page number to 100 then ? or 50 ? or even lower ?
if lower, i just worry on the frontend app infinite scroll on some pages like the list of users or locations where you might want to reach the bottom

config/settings.py
raphodn: fixed here: https://github.com/openfoodfacts/open-prices/pull/1355/commits/95e5f2a2d0bfd125853ea5ca5f42ec7bf8390aaa

open_prices/api/pagination.py
raphodn: fixed here: https://github.com/openfoodfacts/open-prices/pull/1355/commits/95e5f2a2d0bfd125853ea5ca5f42ec7bf8390aaa (used ParseError as it's the only one that returns a 400 and that i can add a custom message)

open_prices/api/pagination.py
raphodn: so which page number limit should we set ?

|max page|size|number of possible results|
|-|-|-|
|1000|10 (default)|10k|
|1000|100 (max)|100k|
|500|10 (default)|5k|
|500|100 (max)|50k|
|100 ?|10 (default)|1k|
|100 ?|100 (max)|10k|

TTalex: I think we can start with 1000 and see, the goal is to catch silly bots.

Slow queries are way beyond that, example with the last one detected :
```
SELECT [...]
FROM "prices"
LEFT OUTER JOIN "products" ON ("prices"."product_id" = "products"."id")
LEFT OUTER JOIN "locations" ON ("prices"."location_id" = "locations"."id")
LEFT OUTER JOIN "proofs" ON ("prices"."proof_id" = "proofs"."id")
ORDER BY "prices"."created" DESC
LIMIT 100
OFFSET 268100
```

```
Query String
size 100
page 2682
location__osm_country dk
order_by -created
```

TTalex (approved):
Didn't run it, but I trust the unit tests :)

open_prices/api/pagination.py
raphodn: changed to 500 :hocho: :pear: 
https://github.com/openfoodfacts/open-prices/pull/1355/commits/922adf7b8e51e6cbac1b1f9b193f200b0f69f3d2

PR #1358, opened 2026-06-19
feat(Proofs): add receipt anonymization

### LINKED ISSUE #957: Allow to anonymize receipts

raphael0202:
This PR adds the receipt anonymization feature.

Remaining tasks to be done:
- [x] Only run the anonymization on draft proofs, and remove the prediction when the draft flag is removed (to preserve privacy)
- [x] Add unit tests to `locate_word` func
- [x] Finish to configure PaddleX on staging, and update `container-deploy.yml` with the service IP address

In future PRs:
- [ ] Remove the Google Cloud OCR integration entirely
- [x] Add a new endpoint that takes the coordinates of the words to be redacted, and overwrite the saved image with redacted parts


open_prices/proofs/models.py
```diff
@@ -535,11 +526,29 @@ def __str__(self):
...
+            type=proof_constants.PROOF_PREDICTION_RECEIPT_ANONYMIZATION_TYPE,
+        ).delete()
+
+
 @receiver(signals.post_save, sender=ProofPrediction)
 def proof_prediction_post_create_increment_counts(sender, instance, created, **kwargs):
     if created:
         if instance.proof_id:
-            Proof.objects.filter(id=instance.proof_id).update(
+            Proof.all_objects.filter(id=instance.proof_id).update(
```
raphael0202, line 551: @raphodn I spotted it while debugging a unit test: we didn't switch from Proof.objects to Proof.all_objects when we added the new manager.

.github/workflows/container-deploy.yml
```diff
@@ -126,7 +129,7 @@ jobs:
           echo "COMPOSE_HTTP_TIMEOUT=180" >> .env
           echo "COMPOSE_PROJECT_NAME=open_prices" >> .env
           echo "COMPOSE_PATH_SEPARATOR=;" >> .env
-          echo "COMPOSE_FILE=docker-compose.yml;docker/prod.yml" >> .env
+          echo "COMPOSE_FILE=docker-compose.yml;docker/prod.yml;docker/ocr.yml" >> .env
```
raphael0202: Remove this change, as we don't use docker in staging/prod.

raphodn (approved):

open_prices/proofs/models.py
```diff
@@ -535,11 +526,29 @@ def __str__(self):
...
+):
+    """Delete the receipt anonymization prediction once the proof is not in
+    draft mode anymore. Receipt anonymization contain PII data, so it's
+    best that those are not publicly available."""
+    if (
+        not created
+        and instance.draft is False
+        and instance.type == proof_constants.TYPE_RECEIPT
+    ):
+        ProofPrediction.objects.filter(
```
raphodn, line 541: possible refacto: move this closer to the draft/non draft logic (view)

PR #1363, opened 2026-06-20
feat(Badges): new `Badge` table to store the underlying rules

raphodn:
first step of #1362

### What

- new `badges` Django app
- new `Badge` model
  - name, description & image_url field
  - rules based on a combination of `metric` (user count field) & `threshold` (the count needed to reach this number)
  - user_count for stats
- admin integration
- basic test

### Screenshot

how it could look like

|List|Detail|
|-|-|
|image|image|

PR #1365, opened 2026-06-21
feat(Badges): new `UserBadge` model to store achievements. Update every night

raphodn:
### What

Following #1363 (new `Badge` model)

- new `UserBadge` model to store the user's achievements
- admin integration
- when is it calculated ?
  - every night via a cron
  - because it's dependent on user counts, we increase the frequency of this cron as well to every night
    - and calculate the badges 5 minutes after
  - the cron also updates the badge's user_count
  - what about real time ? needs #1371 

### Screenshot

image


open_prices/common/tasks.py
```diff
@@ -166,7 +177,8 @@ def history_cleanup_task():
     "challenge_tasks": ("30 1 * * *", {}),  # daily at 01:30
     "history_cleanup_task": ("0 3 * * 1", {}),  # daily at 03:00
     "proof_draft_cleanup_task": ("*/5 * * * *", {}),  # every 5 minutes
-    "update_user_counts_task": ("0 2 * * 1", {}),  # every start of the week (at 02:00)
+    "update_user_counts_task": ("0 2 * * 1", {}),  # daily at 02:00
```
raphodn, line 180: update_user_counts_task: changed from weekly to daily

Issue #1362, opened 2026-06-20
Badge engine


raphodn:
### Story

Following a [survey](https://github.com/openfoodfacts/open-prices/discussions/399#discussioncomment-17191183) to the community regarding gamification, "badges" is part of the top choices.

#### Architecture

- UserBadge table
- BadgeDefinition table ? or write the rules in code to start with ?

```
Projects with badges commonly use one of these patterns:
- Denormalized counters on user/profile rows. <-- our case
- Materialized aggregates in separate stats tables.
- Event logs plus periodic aggregation jobs.
```

#### Levels

ℹ️ ideas, to be discussed :)

see https://github.com/openfoodfacts/open-prices/discussions/399#discussioncomment-17365743

|field|name|steps|delay (currently)|
|-|-|-|-|
|price_count|Contributor|1 / 10 / 100 / 1000 / 10000|live|
|proof_count|Photographer|1 / 10 / 100 / 1000|live|
|location_type_osm_city_count|Tourist|5 / 20 / 50|weekly|
|location_type_osm_country_count|Globetrotter|2 / 5 / 10|weekly|
|challenge_count|Challenger|1 / 5 / 10|weekly|
|year_count|Historian|2 / 5 / 10|weekly|
|proof_type_receipt_count|Shopper|||
|price_in_proof_not_owned_count|Validator||

#### Frontend integration

see https://github.com/openfoodfacts/open-prices-frontend/issues/137#issuecomment-4905693573

### Other open source projects

- Discourse: https://forum.openfoodfacts.org/badges
- BookBrainz: https://bookbrainz.org/editor/23/achievements

raphodn:
## Todo

- [x] Model
  - new `Badge` model: #1363
  - new `UserBadge` table to store achievements & update via nightly cron: #1365
- [x] API
  - endpoint to list all the badges: #1364
  - endpoint to get a specific badge: #1387
  - endpoint to list a user's list of badges: #1366 & #1392
  - endpoint to list a badge's list of users: #1388 & #1391
- [x] Stats
  - User.badge_count: #1378
  - TotalStats.badge_count: #1379

raphodn:
v2 & futur
- Manage complex badges
  - e.g. X prices in a given country
    - tech: materialized view ? jsonfield to store many more counts ?
  - X prices in 1h
- threshold : max ? and replace them ?
- group badges together
- real time ? notification ?
- moderation: cleanup of badges for users who see their prices deleted

PR #1373, opened 2026-06-24
refactor(Tasks): move logic to model classmethods. so that they can be called elsewhere (e.g. tests)

raphodn:
### What

When doing #1362 I had issues calling the `update_badge_task` from the `tasks.py`.
Recommended way is to move the logic to the `Badge` model as a classmethod.

Did this for all models that needed it
- TotalStats
- Product
- User
- Location
- Challenge
- Badge

### DIFF
--- a/open_prices/common/tasks.py
+++ b/open_prices/common/tasks.py
@@ -58,54 +58,27 @@ def import_all_product_db_task():
 
 
 def update_total_stats_task():
-    """
-    Update all total stats
-    """
-    total_stats = TotalStats.get_solo()
-    total_stats.update_price_stats()
-    total_stats.update_product_stats()
-    total_stats.update_location_stats()
-    total_stats.update_proof_stats()
-    total_stats.update_price_tag_stats()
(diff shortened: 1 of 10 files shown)


open_prices/common/tasks.py
raphodn, line 1: the main changes are here: we move the task logic to each model

open_prices/common/tasks.py
```diff
@@ -174,7 +133,7 @@ def history_cleanup_task():
     "update_total_stats_task": ("0 1 * * *", {}),  # daily at 01:00
     "fix_proof_fields_task": ("10 1 * * *", {}),  # daily at 01:10
     "moderation_tasks": ("20 1 * * *", {}),  # daily at 01:20
-    "challenge_tasks": ("30 1 * * *", {}),  # daily at 01:30
+    "update_challenge_task": ("30 1 * * *", {}),  # daily at 01:30
```
raphodn, line 136: renamed for homogenization

Issue #1018, opened 2025-09-25
Locations: manage OSM changes (e.g. a shop changes brand)


raphodn:
### Story

Following #388 we store a Location.osm_version.

We now need a mechanism to actually create different location objects when the same OSM ID is used for shops that have changed.

### Why

- the newly added prices will most likely be linked to the new shop brand, not the old one
- but currently, once we create a location entry for a given osm_id/osm_type, it cannot be modified

### Example

https://www.openstreetmap.org/node/298872652
- version 17 / 2024-07-07 / Intermarché / rebranded!
- version 16 / 2023-07-26 / Casino

### Ideas

- allow having 2 Location with the same osm_id & osm_type. just the osm_version that changed.
- a osm_version would be tied to a osm_version_date ? and prices before or after would go to a different location if it exists ?

Hop-Le133884:
Hi! I'm a student in the CodePath AI301 program and I'd like to work on this issue.
I've read through the codebase and PR #1021 to understand the current exploration scripts. Here's my understanding of what needs to be done:

1. Add osm_version_date field to the Location model to track when each OSM version started
2. Update the UniqueConstraint from (osm_id, osm_type) to (osm_id, osm_type, osm_version) to allow multiple Location records for the same physical store across rebrands
3. Update the script in PR #1021 to actually create the new Location record when a rebrand is detected (currently it only prints)
4. Handle edge cases like stores that have rebranded more than twice by always comparing against the latest osm_version per osm_id + osm_type and checking for duplicates before creating

Old Location records and their prices would be kept intact,  new prices would route to the new Location based on osm_version_date.
I'll start by reproducing the problem locally this week. Any pointers on where to start in the codebase would be appreciated!

raphodn:
Hi @Hop-Le133884 sorry for the super late reply. I had a look at your PR, and I'm picking the pieces slowly back together.

I'm going to start by merging a PR just for 1. (`osm_version_date`).

raphodn:
I merged #1021 and ran it in staging. It's an "exploration script" to see how many locations need updating, and why.
Things i see pop up quite a bit: 
- minor changes without incidence
- name & brand change
- POIs deleted
- POIs marked as disused
- a supermarket that was tagged as a `way`, and it is then stripped + a `node` is created next to it (Biedronka ^^)

raphodn:
I also merged #1435 which starts to auto-update locations with "minor" updates
- it's not yet an auto-update actually, it requires to run the script
[...]

PR #1374, opened 2026-06-25
feat(Locations): support OSM location versioning (rebrand detection)

Hop-Le133884:
### What
 - Add `osm_version_date` field to `Location` model to track when each OSM version became active
 - Update unique constraint from `(osm_id, osm_type)` to `(osm_id, osm_type, osm_version)` to allow multiple versions of the same physical location to coexist
 - Update `get_location_from_openstreetmap()` to return 'version_date' from OSM API response
 - Update `fetch_and_save_data_from_openstreetmap()` to detect meaningful rebrands (version changed + name or brand changed) and create a new Location record instead of overwriting
 - Extend `set_location_osm_brand_and_version` backfill command to populate `osm_version_date` for existing records
 - Add `LocationVersioningTest` and update existing unique constraint test

#### Before FIX:
image
#### after FIX:
image

### Fixes bug(s)
- #1018 

### Part of 
- #1018 

### How to test
```bash
docker compose -f docker-compose.yml -f docker/dev.yml exec api uv run python manage.py test open_prices.locations --verbosity=2 2>&1 | tail -15
# 16 tests should pass as showed in the Screenshot above
```
### Notes
Build on the detection logic from draft PR #1021 by raphodn. The change detection uses `version changed AND (name or brand changed)` to distinguish meaningful rebrands from minor OSM edits like opening hours updates.


TTalex:
Hey,

In the Open Prices process, when new proofs are uploaded with some location information (`location_osm_id` and `location_osm_type`), location is fetched or created :
https://github.com/openfoodfacts/open-prices/blob/3810597b0a035d145f480e01391c7b3c7b19e482/open_prices/proofs/models.py#L282-L292

Reading this PR, I'm not too sure what would happen when multiple location objects exist for identical `location_osm_id` & `location_osm_type`. Would it fetch the latest one ?

Hop-Le133884:
Good catch! I added a `get_latest_by_osm()` QuerySet method that returns the most recent version  (ordered by osm_version desc, then updated desc), if not then create new. Updated two call sites:

1. `proofs/models.py, set_location()` - replaced `get_or_create()` with `get_latest_by_osm()` + create-if-missing. Preserves the same behavior but handles the multi-version case without throwing `MultipleObjectsReturned`.
2. `api/locations/views.py, create()` duplicate detection now checks `get_latest_by_osm()` before saving instead of catching a constraint error after. This also fixes a CI test failure (`test_location_create_duplicate`) caused by `NULL != NULL` semantics in the new unique constraint when osm_version is not set.

Fetching now by osm_id + osm_type always returns the current tenant of that physical OSM node. Older versions remain accessible via the location's id for historical prices linkage.

open_prices/locations/tasks.py
```diff
@@ -5,6 +5,29 @@
 def fetch_and_save_data_from_openstreetmap(location: Location):
```
raphodn, line 5: this method is called on location post_create (post_save + created=True), so isn't it too late to check for changes ?

raphodn: ok nevermind i see what you did, where calling the `get_latest_by_osm` avoids to create a new location as the osm_version would be empty for the new one and the constraint wouldn't pop up

raphodn: my concern is what is a "meaningful" change. and probably put a human in the loop to re-merge locations afterwards  ? (e.g. name change because of typo, not a rebranding)

raphodn: just that `fetch_and_save_data_from_openstreetmap` is currently called only on Location creation (post_save + created=True). so I don't see when it will be able to update an existing location. or we need to make more changes.

PR #1375, opened 2026-07-02
feat: add new route to anonymize receipt

raphael0202:
Add an API to anonymize draft receipts.
The front-end should fetch the RECEIPT_ANONYMIZATION prediction, display it to the user (who can optionally modify the areas to redact), and send the results back to this new route.
The route modifies the image in place on disk.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -196,7 +196,7 @@ CORS_ALLOW_CREDENTIALS = True
 # https://pillow.readthedocs.io/
 # ------------------------------------------------------------------------------
 
-THUMBNAIL_SIZE = (400, 400)
+THUMBNAIL_SIZE = 400
 
 
 # Django REST Framework (DRF) & django-filters & drf-spectacular
--- a/open_prices/api/proofs/views.py
+++ b/open_prices/api/proofs/views.py
@@ -297,7 +297,9 @@ class ProofDraftViewSet(
         Only the file and type fields are expected. ML models runs immediately asynchronously."""
         return base_upload(request, draft=True)
 
-    @extend_schema(request=DraftProofAnonymizeRequestSerializer)
+    @extend_schema(
+        request=DraftProofAnonymizeRequestSerializer, responses=ProofFullSerializer
+    )
     @action(
         detail=True,
         methods=["POST"],
@@ -341,7 +343,7 @@ class ProofDraftViewSet(
             proof._change_reason = "Saving new image path after anonymization"
             proof.save()
 
-        return Response(status=status.HTTP_200_OK)
+        return Response(ProofFullSerializer(proof).data, status=status.HTTP_200_OK)
 
     @extend_schema(request=ProofUpdateSerializer, responses=ProofFullSerializer)
     def partial_update(self, request: Request, pk=None) -> Response:
--- a/open_prices/api/proofs/views.py
+++ b/open_prices/api/proofs/views.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 import PIL.Image
 from django.conf import settings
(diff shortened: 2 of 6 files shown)


config/settings.py
```diff
@@ -196,7 +196,7 @@
 # https://pillow.readthedocs.io/
 # ------------------------------------------------------------------------------
 
-THUMBNAIL_SIZE = (400, 400)
```
raphael0202, line 199: The function to generate thumbnail using OpenCV (generate_image_thumbnail_cv2) takes a single `max_size` argument, so I updated the `generate_thumbnail` function to use the same parameter as default value.

open_prices/proofs/utils.py
```diff
@@ -377,7 +443,9 @@ def generate_image_thumbnail_cv2(image: np.ndarray, max_size: int) -> np.ndarray
         scaling_factor = max_size / max(width, height)
         new_width = int(width * scaling_factor)
         new_height = int(height * scaling_factor)
-        image = cv2.resize(image, (new_width, new_height), interpolation=cv2.INTER_AREA)
+        image = cv2.resize(
+            image, (new_width, new_height), interpolation=cv2.INTER_LANCZOS4
```
raphael0202, line 447: Using here Lanczos algorithm for interpolation, to be consistent with Pillow-based `generate_thumbnail` function.

open_prices/api/proofs/views.py
```diff
@@ -289,6 +297,52 @@ def upload(self, request: Request) -> Response:
...
+            bounding_boxes=serializer.data["bounding_boxes"],
+        )
+
+        if new_image_path != image_path or new_image_thumb_path != image_thumb_path:
+            proof.file_path = str(new_image_path)
+            proof.image_thumb_path = str(new_image_thumb_path)
+            proof._change_reason = "Saving new image path after anonymization"
+            proof.save()
+
+        return Response(status=status.HTTP_200_OK)
```
TTalex: Could be nice to send some info back (like the proof object ?), especially if file_path might have changed.

open_prices/api/proofs/views.py
```diff
@@ -289,6 +297,52 @@ def upload(self, request: Request) -> Response:
...
+                {"detail": "Only receipts can be anonymized"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        # Warning: bounding boxes here have the same format [x_min, y_min, x_max, y_max]
+        # as the `Word` bounding boxes in the OCR results.
+        # Currently, the format used by the bounding boxes of price tags use a different
+        # format ([y_min, x_min, y_max, x_max]).
+        # The [x_min, y_min, x_max, y_max] format is much more common, and we should migrate
+        # price tag bounding boxes to that format in the future for consistency.
```
TTalex, line 324: Sad news. does that mean that our two bounding box detection models are replying different things ?

TTalex (approved):

open_prices/api/proofs/views.py
raphael0202: Good catch, indeed the front-end may need the new image path. I'm adding the full proof as response.

open_prices/api/proofs/views.py
raphael0202: PaddleOCR returns bounding box in the `[x_min, y_min, x_max, y_max]` format.
The bounding box format of Yolov8 models (the architecture we use) is `[x_center, y_center, box_width, box_height]`, and in the openfoodfacts-python lib, we transform it into `[y_min, x_min, y_max, x_max]`.

`[y_min, x_min, y_max, x_max]` is used on Robotoff, as it's the format used by Google Cloud Vision (OCR), but it's not common outside this.

PR #1376, opened 2026-07-05
fix: remove unreachable elif in barcode_fix_short_codes_from_usa

### LINKED ISSUE #1367: Logic error in openfoodfacts.py
### What

I find one logic error in code.

Image

The elif len(barcode) == 12 will never run, because 12 is already handled by the first if.
### Platform (Desktop, Mobile, Hunger Games)
 - OS:  windows
 - Platform  desktop

### Part of
-

munzzyy:
Fixes #1367.

`barcode_fix_short_codes_from_usa` had a dead branch: the first condition already catches 10, 11, and 12-digit barcodes (`if len(barcode) in (10, 11, 12)`), so the `elif len(barcode) == 12` right after it could never run.

I didn't just delete it though, I checked whether the elif's own logic (prepend a single `"0"`) was actually the correct behavior for 12-digit codes and the fix should have made it reachable instead. It isn't: the existing test case `("471058718081", "4710587180816")` only works through the pad-and-check-digit branch. Prepending `"0"` gives `0471058718081`, which doesn't match. So removing the dead code is the right fix, not resurrecting it.

Also updated the docstring, since it still described the "12 digits: add a leading zero" behavior that the dead branch used to represent.

Verification: couldn't run the full Django test suite locally (needs Postgres + `.env`), but I installed the `openfoodfacts` package standalone and ran the fixed function's logic against all five cases from `test_barcode_fix_short_codes_from_usa` directly against the real `calculate_check_digit`/`has_valid_check_digit` implementations, all pass. Also ran `ruff check` and `ruff format --diff` against the touched file, both clean.

### DIFF
--- a/open_prices/common/openfoodfacts.py
+++ b/open_prices/common/openfoodfacts.py
@@ -395,8 +395,7 @@ def barcode_fix_short_codes_from_usa(barcode: str) -> str:
     """
     Fix short barcodes from USA
 
-    10 or 11 digits: pad them to 12 digits and calculate their check digit
-    12 digits: add a leading zero
+    10, 11 or 12 digits: pad them to 12 digits and calculate their check digit
 
     This function is based on the fact that most of the short barcodes are from
     the USA, and that they can be converted to a valid EAN-13 barcode by
@@ -413,10 +412,6 @@ def barcode_fix_short_codes_from_usa(barcode: str) -> str:
         barcode_temp = barcode.zfill(12)
         barcode_check_digit = openfoodfacts.barcode.calculate_check_digit(barcode + "0")
         barcode_temp = barcode_temp + barcode_check_digit
-        if barcode_is_valid(barcode_temp):
-            return barcode_temp
-    elif len(barcode) == 12:
-        barcode_temp = "0" + barcode
         if barcode_is_valid(barcode_temp):
             return barcode_temp
     return barcode


raphodn (approved):

open_prices/common/openfoodfacts.py
```diff
@@ -395,8 +395,7 @@ def barcode_fix_short_codes_from_usa(barcode: str) -> str:
     """
     Fix short barcodes from USA
 
-    10 or 11 digits: pad them to 12 digits and calculate their check digit
-    12 digits: add a leading zero
+    10, 11 or 12 digits: pad them to 12 digits and calculate their check digit
```
raphodn, line 398: @TTalex i need your help actually, i'm a bit lost
after re-reading [this PR](https://github.com/openfoodfacts/open-prices/pull/1022), but also this [Slack thread](https://openfoodfacts.slack.com/archives/C02Q6KM7F/p1758573105433139)
when we have a 12-digit barcode, should we pad it with 0, or calculate its check digit ? ^^

raphodn: https://en.wikipedia.org/wiki/International_Article_Number

> Note that EAN-13 codes beginning with 0 are actually 12-digit UPC codes with a prepended 0 digit

raphodn: though for example this proof: https://prices.openfoodfacts.org/proofs/48849
471058718125 (12) was transformed into 4710587181257 (13)

TTalex: From my understanding, US barcodes:

- Do not show leading zeros
- Do not show checksum

So we have to add both:

- First pad leading zeros, if needed, until we reach 12 digits
- Then compute checksum

For sanity checking, the [first three numbers of EAN13 are the country of origin](https://www.barcode-us.com/tools-and-resources/ean-country-code-table), in your example, "471" is Taiwan, which matches the product showing "made in Taiwan".

Another example:
| What can be printed (10 digits) | What can be printed (11 digits) | Pad to 12  | Add check digit to EAN-13  |
| --- | --- | --- | --- |
| 4139000105  | 04139000105 | 004139000105  | 0041390001055 |

The Wikipedia article is a bit misleading, because it assumes 12-digit UPC with check digit included. In that case, adding a leading zero works. 
In your example, 471058718125 is not a valid 12-digit UPC, 5 not being the check digit of 47105871812.

Now the question is, can we have both cases printed in real life ?
* 12 digit codes without check digit, thus missing it at the end
* 12 digit codes with check digit, thus missing leading zero

My guess is we cannot have the latter, but I have no proof.

---

Maybe our test should do the opposite, for 12 digits codes:
* First, add leading zero and test checksum validity
* If fails, add check digit at the end

But I'm sure we can find outliers that randomly match the first checksum validity, example (not a real product) : 611210005473 (12)
* 0611210005473 (13) is a valid code for a product made in the US or Canada
* 6112100054730 (13) is a valid code for a product made in Morocco

munzzyy: Checked this against your conclusion, and the code that's left after removing the elif already does exactly this: pad to 12, then compute the checksum.

open_prices/common/openfoodfacts.py:411-415, unified now for 10/11/12-digit inputs:

    barcode_temp = barcode.zfill(12)
    barcode_check_digit = openfoodfacts.barcode.calculate_check_digit(barcode + "0")
    barcode_temp = barcode_temp + barcode_check_digit

It doesn't call zfill before computing the check digit, but that ends up being the same result either way. calculate_check_digit weighs digits by their position counting from the right, so any zeros padded onto the left don't change the sum. Ran it on your 471058718081 example: check digit comes out to 6, so 4710587180816, country code 471, Taiwan, same as what you found.

The elif this PR removes did the other thing, just prepending "0" with no check digit, which for that same barcode gives 0471058718081 (country code 047, Belgium/Luxembourg) - the wrong path, and unreachable anyway since the if above already catches length 12.

So no behavior changes here, just the dead branch coming out and the docstring updated to match what the surviving code actually does.

TTalex: Yeah, ok, let's merge.

We can always add some kind of prepending 0 logic later if we notice errors on UPC12 codes with checksums.

PR #1377, opened 2026-07-07
feat(Prices): add outlier detection algorithm

### LINKED ISSUE #934: Detect price outliers
Detect possible erroneous prices (outliers): prices that are too high/too low for the product_code/category_tag.

We would store this information in the same way as prices to moderate.

raphael0202:
Implementation of a price outlier detection algorithm.

Regarding its integration, I was considering the following:

- Every morning (after midnight), fetch price outliers (`find_outliers`) with a target date set to the day before.
- Create new flags in the moderation table corresponding to each outlier detected.
- Refresh the `price_statistics_5y` materialized view

The advantage of refreshing the view *after* running the outlier detection algorithm is that the new price is not part of the group statistics.


TTalex:
We might need to be a bit less strict on the acceptable range. 3*stdev seems quite small.
Example, taken randomly : https://prices.openfoodfacts.org/products/3450970163675
4th price (4.61€) would have triggered an error, 5th price (4.57€) as well, 7th price (4,55 €) as well

At least avoid alerts when stdev is very small, or 0.

Also, grouping should be done on discounts.

raphael0202:
The doc was outdated, sorry for that: we now classify a price as an outlier when it's either > 3 * med or < med / 3.
I first started with the stddev, and came to the conclusion than using the median gave better results.

> Also, grouping should be done on discounts.

I don't think it's necessary, I tested with production data and the outliers I checked were not discounted prices.
I think the price variations due to discounts are not big enough to turn discounted prices into outliers.

docs/topics/outlier-detection.md
raphodn, line 1: cool ! but to be visible by mkdocs (i think ?), you need to add it to `mkdocs.yml`

docs/topics/outlier-detection.md
```diff
@@ -0,0 +1,89 @@
...
+- type (`PRODUCT` or `CATEGORY`)
+- country code (ex: `FR`)
+- price per (`UNIT` or `KILOGRAM`)
+- currency code (ex: `EUR`)
+
+and compute the price median for each group.
+
+Let's call `x` the price we want to run the outlier detection algorithm against and `med` the price median of the group associated with the price.
+
+If there at least 3 prices in the group, and `x` is higher than  `3 * med` or lower than `med / 3`, we classify the price as an outlier.
```
raphodn:
```suggestion
If there are at least 3 prices in the group, and `x` is higher than  `3 * med` or lower than `med / 3`, we classify the price as an outlier.
```

open_prices/prices/migrations/0017_add_price_statistics_materialized_view.py
```diff
@@ -0,0 +1,75 @@
...
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("prices", "0016_alter_historicalprice_owner_alter_price_owner"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
```
raphodn, line 12: i'm wondering if this should be in a migration, or in the code instead.
for instance in the refresh_materialized_view it could check if the view exists, and create it if not.
because if we change often, or add new views, it'll create a lot of migrations 🤔

raphodn: it's a first for me to have a materialized view in Django.

but i guess it's like a model, so indeed a migration is best suited 👍

open_prices/prices/outlier_detection.py
raphodn, line 1: idea: move this to the `moderation` folder ?

open_prices/prices/models.py
```diff
@@ -530,3 +530,27 @@ def price_post_delete_update_duplicate_of(sender, instance, **kwargs):
     for price in Price.objects.filter(duplicate_of_id=instance.id).order_by("-created"):
         price.duplicate_of = None
         price.save(update_fields=["duplicate_of"])
+
+
+class PriceStatistics5y(models.Model):
```
raphodn, line 535: where is this model used ? except to call refresh
could it be used in `find_outliers` to build the query instead of RAW SQL ?

open_prices/prices/tests.py
```diff
@@ -1231,3 +1233,149 @@ def test_normalize_barcodes_command(self):
...
             self.assertEqual(price.product_code, "0123456789100")
             self.assertEqual(price.product_id, Product.objects.first().id)
+
+
+class TestOutlierDetection(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.product_code = "0123456789100"
+        cls.category_tag = "en:apples"
+        cls.product = ProductFactory.create(code=cls.product_code)
```
raphodn:
```suggestion
        cls.product = ProductFactory(code=cls.product_code)
```

no need to call create. you can call the class directly (you can look at the other tests we never use it)
https://factoryboy.readthedocs.io/en/stable/reference.html#factory.Factory.__call__

open_prices/prices/tests.py
```diff
@@ -1231,3 +1233,149 @@ def test_normalize_barcodes_command(self):
...
+                if not k.startswith("price_per")
+            },
+            price="250",
+            price_per=price_constants.PRICE_PER_UNIT,
+        )
+        PriceStatistics5y.refresh_materialized_view()
+
+    def test_find_outliers(self):
+        price_outliers = list(find_outliers(median_threshold=3, min_count=3))
+        product_outliers = [
```
raphodn:
```suggestion
        price_product_outliers = [
```
to clarify, i thought for a sec we actually had product outliers ^^

raphodn (approved):

open_prices/prices/models.py
raphodn: had a chat with Qwen and the SQL seems the most optimal way :eyes:

docs/topics/outlier-detection.md
raphael0202: You don't need to, you only need to add it if you want to change the default name of the document.

open_prices/prices/outlier_detection.py
raphael0202: I was planning to call the find_outliers from the moderation app, so that concerns are well splitted.
I prefer having the price outlier detection code in the price app, but I'm fine to moving it to moderation.

PR #1402, opened 2026-07-31
fix: add permissions to top-issues GitHub Actions

teolemon:
fix: add permissions to top-issues GitHub Actions

- `contents: read` (read repository metadata/content)
- `issues: write` (label/update issues and dashboard issue/comment if used)
- `pull-requests: write` (the action can include top pull requests and may update PR-related artifacts)

### DIFF
--- a/.github/workflows/top-issues.yml
+++ b/.github/workflows/top-issues.yml
@@ -1,4 +1,5 @@
-name: Top issues action.
+name: Top issues action
+
 on:
   schedule:
   - cron:  '0 0 */1 * *'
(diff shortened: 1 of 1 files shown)


.github/workflows/top-issues.yml
```diff
@@ -14,6 +14,10 @@ jobs:
     name: Display and label top issues.
     runs-on: ubuntu-slim
     if: github.repository_owner == 'openfoodfacts'
+    permissions:
```
raphodn: why is it being added inside the job and not outside like others ?

teolemon: both are permissible per the doc: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions

raphodn: ok but if there is 1 job, i'd prefer to have all the actions follow the same pattern and have them at the top then, much more readable

raphodn (approved):

.github/workflows/top-issues.yml
raphodn: updated here: https://github.com/openfoodfacts/open-prices/pull/1402/commits/94f2112976af1d4f04e05b3303849e8cef7d317c

PR #1419, opened 2026-08-15
feat(Prices): API: new geo filter (lat, lon & radius_km)

### LINKED ISSUE #277: Prices API: new filter for locations within n meters, after providing latitude/longitude
---
### Problem
By @monsieurtanuki 

### Proposed solution
@teolemon Not really from the PoC, but it would be wonderful if we could filter prices for locations within n meters, after providing latitude/longitude:

- we could know the prices in shops that are close enough to me (instead of remote cheaper / more expensive locations)
- we could imagine a crazy feature like "this is my shopping cart: tell me which shop in the neighborhood is the cheapest"
- without the lat/lon/meter filter, providing a list of locations (instead of a single) - give me the prices for "my favorite locations"

Besides, the most annoying part is finding the location. Not convinced so far by what I implemented in the PoC or what is implemented in prices. Still looking for a "get the shops that match 'monoprix avenue daumesnil paris'" easy method.

### Additional context

_Originally posted by @monsieurtanuki in https://github.com/openfoodfacts/smooth-app/issues/5072#issuecomment-1949126838_

### Part of
- #

TaciteOFF:
Closes #277, open since April 2024 and labelled `top issue`.

`GET /prices` now accepts `lat`, `lon` and `radius_km`. It returns only the prices whose location sits inside that radius.

> A second PR follows on `open-prices-frontend`: openfoodfacts/open-prices-frontend#2332. It adds a "nearby" filter to the price list and to product pages, and it needs this endpoint. It stays a draft until this one is merged and deployed.

### Why I did not do this from the client

Since #1303 added `/locations/nearby`, a client can try to combine the two endpoints: fetch the nearby locations, collect their ids, then call `GET /prices?location_id__in=<ids>`. I built exactly that in the frontend. It does not hold up.

**The URL gets too long.** Measured on production, with the params the app really sends:

| Search | Locations | Request line | Result |
|---|---|---|---|
| Paris, 5 km | 407 | 2 833 B | 200 |
| Paris, 10 km | 578 | 3 983 B | 200 |
| Paris, 20 km | 671 | 4 613 B | **400 — `Request Line is too large (4562 > 4094)`** |

**Most of the download is wasted.** For Paris / 10 km the client downloads 459 KB and keeps 2 491 bytes of ids, so 0.5 % of it. `/locations/nearby` returns 26 fields per location when only `id` is needed, and `size` is capped at 100.

**Loading the locations in batches does not help either.** `order_by` would then apply inside each batch instead of the whole result. On Paris / 10 km sorted by `-created`, the 10 most recent prices of the area all come from locations outside the 20 nearest ones. The list would start one month in the past, then jump forward while the user scrolls. And the total would show 833 instead of 78 042.

### How it works

`PriceFilter.filter_queryset()` runs a subquery on the existing `Location.objects.nearby()`, so `/prices` and `/locations/nearby` share one definition of "nearby". `filters.py` imports `Location` for that.

`lat`, `lon` and `radius_km` are declared as regular filters, so they show up in the OpenAPI schema with their bounds, but they do not filter on their own: a form-level `clean()` requires the three to be sent together, otherwise the request is rejected with a 400.

That check matters more than it looks. django-filter ignores unknown params, so today `GET /prices?lat=48.85&lon=2.35&radius_km=5` returns 200 with every price, exactly like a request carrying a random param name. Sending `?lat=48.85` alone behaves the same way. Any client that guesses the param names gets a full unfiltered list instead of an error. That is the trap I fell into, and I only caught it late. The labels now say the three go together, so the generated docs show it too.

### It also reaches /prices/stats

`/prices/stats` runs the same `filter_queryset`, so it narrows down as well, and the whole aggregate is computed on the nearby subset. On my local copy of the production data, Paris / 5 km gives `price__count: 21540`, matching the list endpoint. This comes for free and looks useful, so I covered it with tests rather than leave it undocumented.

### Tests

The suite passes. Note that 4 tests around challenges and user badges currently fail on `main` too, independently of this PR — they look date-sensitive and started failing when the date rolled over. I checked by stashing my changes and running them again on a clean tree.

I also loaded the production dumps into a local database (6 783 locations, 114 869 proofs, 290 792 prices) and compared the endpoint against a haversine computed separately over the raw JSONL. 4 cities, 4 radii, **16 exact matches out of 16**:

| | 1 km | 5 km | 20 km | 100 km |
|---|---|---|---|---|
| Paris | 1 214 | 21 540 | 79 545 | 80 078 |
| Lyon | 570 | 11 795 | 16 293 | 63 719 |
| Berlin | 24 | 311 | 1 756 | 4 875 |
| Marseille | 11 | 175 | 202 | 893 |

I checked pagination over the 13 pages of Paris / 1 km: 1 214 distinct ids, no duplicate, `-created` order kept. Combining with `price__gte`, `currency`, `type` and `date__gte` gives a strict subset every time.

**It is faster than the unfiltered query**, on 290 792 prices: 23 ms at 1 km, 81 ms at 5 km, 218 ms at 20 km, against 228 ms with no filter. Even a radius large enough to cover the whole planet answers in ~75 ms for the first page. So there is no need for a maximum radius, and no need for an index on `(osm_lat, osm_lon)` at this table size.

### Possible follow-ups, not in this PR

- **Sorting by distance**, as discussed in review. It works and costs ~55 ms at 5 km, but the distance belongs to the location, not to the price: in Paris within 5 km, 21 540 prices sit on 407 locations, so about 53 prices share each distance value. I would add it as an `order_by=distance` option rather than as the default.
- **The antimeridian.** `Location.objects.nearby()` compares `osm_lon` to `center_lon ± delta` without wrapping at ±180°, so a location at 179.9° and one at −179.9° are about 22 km apart but never match each other. I reproduced it on both sides, and the same longitude gap around 0° works fine. It comes from #1303 and is not introduced here. There is no data in that area today.

### Note for reviewers

`manage.py makemigrations --check` already fails on `main`: `PriceStatistics5y` in `prices/models.py` has no migration. This PR does not touch any model, so that failure does not come from it.

### Two things I found in the published dumps

- 4 proofs have a **negative `price_count`**, down to −9. The schema itself rejects this, since the column is a `PositiveIntegerField`. It looks like a counter going below zero when prices are deleted.
- There is no `products.jsonl.gz` (404), so `product_id` cannot be restored from the dumps.

### DIFF
--- a/open_prices/api/prices/filters.py
+++ b/open_prices/api/prices/filters.py
@@ -1,11 +1,35 @@
 import django_filters
+from django import forms
 
 from open_prices.api.utils import ArrayFieldElementContainsFilter
 from open_prices.common import constants
 from open_prices.locations import constants as location_constants
+from open_prices.locations.models import Location
 from open_prices.prices.models import Price
 from open_prices.products import constants as product_constants
 from open_prices.proofs import constants as proof_constants
+
+NEARBY_PARAMETER_NAMES = ("lat", "lon", "radius_km")
+
+
+class PriceFilterForm(forms.Form):
+    def clean(self):
+        cleaned_data = super().clean()
+        provided_parameters = {
+            name for name in NEARBY_PARAMETER_NAMES if name in self.data
+        }
+        has_empty_parameter = any(
+            self.data.get(name) in (None, "") for name in provided_parameters
+        )
+        if provided_parameters and (
+            len(provided_parameters) != len(NEARBY_PARAMETER_NAMES)
+            or has_empty_parameter
+        ):
+            raise forms.ValidationError(
+                "lat, lon and radius_km must be provided together with non-empty "
+                "values."
+            )
+        return cleaned_data
 
 
 class PriceFilter(django_filters.FilterSet):
@@ -13,6 +37,26 @@ class PriceFilter(django_filters.FilterSet):
     PriceViewSet GET queryset has select_related on product, location, proof
     """
 
+    lat = django_filters.NumberFilter(
+        method="ignore_nearby_parameter",
+        min_value=-90,
+        max_value=90,
+        label="Latitude of the center point (-90 to 90). Must be sent together "
+        "with lon and radius_km.",
+    )
+    lon = django_filters.NumberFilter(
+        method="ignore_nearby_parameter",
+        min_value=-180,
+        max_value=180,
+        label="Longitude of the center point (-180 to 180). Must be sent together "
+        "with lat and radius_km.",
+    )
+    radius_km = django_filters.NumberFilter(
+        method="ignore_nearby_parameter",
+        min_value=0,
+        label="Search radius in kilometers (0 or more). Must be sent together "
+        "with lat and lon.",
+    )
     kind = django_filters.ChoiceFilter(
         choices=constants.KIND_CHOICES,
         method="filter_kind",
@@ -53,8 +97,32 @@ class PriceFilter(django_filters.FilterSet):
             return queryset.has_kind_consumption()
         return queryset
 
+    def ignore_nearby_parameter(self, queryset, name, value):
+        """Validate nearby parameters without applying them separately."""
+        return queryset
+
+    def filter_queryset(self, queryset):
+        queryset = super().filter_queryset(queryset)
+        nearby_parameters = {
+            name: self.form.cleaned_data.get(name) for name in NEARBY_PARAMETER_NAMES
+        }
+        if not all(value is not None for value in nearby_parameters.values()):
+            return queryset
+
+        nearby_location_ids = (
+            Location.objects.nearby(
+                center_lat=float(nearby_parameters["lat"]),
+                center_lon=float(nearby_parameters["lon"]),
+                radius_km=float(nearby_parameters["radius_km"]),
+            )
+            .order_by()
+            .values("id")
+        )
+        return queryset.filter(location_id__in=nearby_location_ids)
+
     class Meta:
         model = Price
+        form = PriceFilterForm
         fields = {
             "type": ["exact"],
             "product_code": ["exact", "in", "isnull"],
--- a/open_prices/api/prices/tests.py
+++ b/open_prices/api/prices/tests.py
@@ -509,6 +509,153 @@ class PriceListFilterApiTest(TestCase):
         self.assertEqual(response.data["total"], 0)
 
 
+class PriceListNearbyFilterApiTest(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.url = reverse("api:prices-list")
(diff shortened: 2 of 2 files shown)


open_prices/api/locations/serializers.py
```diff
@@ -53,7 +54,9 @@ class LocationNearbySerializer(LocationSerializer):
 class LocationNearbyParamsSerializer(serializers.Serializer):
     lat = serializers.FloatField(min_value=-90, max_value=90, required=True)
     lon = serializers.FloatField(min_value=-180, max_value=180, required=True)
-    radius_km = serializers.FloatField(min_value=0, required=True)
+    radius_km = serializers.FloatField(
```
raphodn: this should be in another PR

raphodn: in any case, the result is paginated, so having a max isn't all that mandatory i'd say.

open_prices/api/prices/filters.py
```diff
@@ -53,8 +99,32 @@ def filter_kind(self, queryset, name, value):
             return queryset.has_kind_consumption()
         return queryset
 
+    def ignore_nearby_parameter(self, queryset, name, value):
+        """Validate nearby parameters without applying them separately."""
+        return queryset
+
+    def filter_queryset(self, queryset):
```
raphodn, line 104: if the filtering is done by geo, then you'd expect the prices to be ordered by closest as well no ?

open_prices/api/locations/serializers.py
TaciteOFF: You're right, I'll remove it. I tested it on a local copy of the production data (6.8k locations, 290k prices). Even a radius big enough to cover the whole planet takes about 75 ms for the first page, and about 30 ms for the count.  I'll remove it from /prices too, not only from /locations/nearby. (a limit can always be added later anyway)

open_prices/api/prices/filters.py
TaciteOFF: Good idea, and it works. I tried it: I compute the distance on the price query, from the location's osm_lat and osm_lon. It takes about 55 ms at 5 km and 184 ms at 20 km. Sorting by -created takes about 29 ms and 55 ms. So it is a bit slower, but fine.
 One thing to know: the distance comes from the location, not from the price. In Paris within 5 km, there are 21 540 prices spread over only 407 locations. So about 53 prices share the same distance. You would get all the prices of the closest shop, then all the prices of the next one. It also needs a second sort key, otherwise the pages are not stable.
So I would add it as an option, order_by=distance, instead of making it the default. Someone looking for prices near them often wants the cheapest or the newest first.

raphodn (approved):

raphodn:
Thanks a lot :+1:

PR #1425, opened 2026-08-29
fix: update stale hardcoded category taxonomy ids in RawCategory (ml)

Eduardo-Vilas-Boas:
Fixes #1398, related to #1338.

`RawCategory` (`open_prices/proofs/ml/common.py`) is a hardcoded enum used as the
Gemini structured-output schema constraint for `Label.category` in
`open_prices/proofs/ml/price_tags.py`. When Open Food Facts renames a category's
canonical taxonomy id, Gemini can only ever return whatever string is still listed
in this enum — it has no way to output the new canonical id, even when the old one
is no longer canonical. This is what #1398 refers to as "the list of bulk products
being hard coded": the AI keeps suggesting `en:kiwis` because that was the only
kiwi-shaped enum value available, not `en:kiwifruits`.

This PR re-syncs `RawCategory` with the current OFF category taxonomy. Auditing
all 83 values in the enum against `taxonomies/food/categories.txt` turned up 5
that had drifted from their current canonical id:

| enum had | canonical is now |
|---|---|
| `en:kiwis` | `en:kiwifruits` |
| `en:mandarin-oranges` | `en:mandarins` |
| `en:garlic` | `en:garlics` |
| `en:pineapple` | `en:pineapples` |
| `en:rhubarbs` | `en:rhubarb` |

All other values matched their current canonical id — no further drift found.

## Evidence

Canonical names are defined positionally in `taxonomies/food/categories.txt`: on a
`lang: Name1, Name2, ...` line, the first name is canonical for that language: the
rest are synonyms. Source: [openfoodfacts-server@085f0527](https://github.com/openfoodfacts/openfoodfacts-server/blob/085f0527b5054cbedef246483f961360a891888f/taxonomies/food/categories.txt).

**Kiwi** — [L70329-L70330](https://github.com/openfoodfacts/openfoodfacts-server/blob/085f0527b5054cbedef246483f961360a891888f/taxonomies/food/categories.txt#L70329-L70330) (see also openfoodfacts-server#13821, which standardized on "kiwifruit"):
```
< en: Fruits
en: Kiwifruits, Kiwis, Kiwi fruits
```

**Mandarin** — [L69652-L69653](https://github.com/openfoodfacts/openfoodfacts-server/blob/085f0527b5054cbedef246483f961360a891888f/taxonomies/food/categories.txt#L69652-L69653):
```
< en: Citrus
en: Mandarins, Mandarines, Mandarin oranges
```

[...]

Issue #1398, opened 2026-07-28
Update ml bulk category detection for kiwifruits


TTalex:
https://github.com/openfoodfacts/openfoodfacts-server/pull/13821 changed the category canonical name from en:kiwis to en: kiwifruits

Our AI process still suggests en:kiwis, the list of bulk products being hard coded

TTalex:
Related https://github.com/openfoodfacts/open-prices/issues/1338

TTalex:
Same for mandarins, I believe

PR #1426, opened 2026-08-30
fix: re-normalize price category tags when the taxonomy renames a canonical id

### LINKED ISSUE #1338: Old categories don’t (always?) get merged/renamed with taxonomy updates
### What
In some cases (which might themselves be issues that need fixing), users can add bulk items to categories that don’t currently have an English name but later gets an English name. This leaves the bulk items separate from the (now canonical) English name.

1. Actually create the products in a relevant (non-?)category. I don’t know how to do this, so I’m using https://prices.openfoodfacts.org/products/de:rumkugeln as a point of reference.
2. Note that a category taxonomy lookup on [`de:rumkugeln`](https://world.openfoodfacts.org/facets/categories/de:rumkugeln) returns `en:rum-balls` and compare with https://prices.openfoodfacts.org/products/en:rum-balls
3. Note that the two pages do not contain the same prices/products

### Expected behavior
All entries of `de:rumkugeln` should appear with/under `en:rum-balls` (and probably the `de:rumkugeln` URL should redirect to `en:rum-balls`: https://github.com/openfoodfacts/open-prices-frontend/issues/2227).

### Screenshots

de:rumkugeln with 2 prices

en:rum-balls with 1 price

Eduardo-Vilas-Boas:
Fixes #1338. Related to #1398 (already fixed separately, on https://github.com/openfoodfacts/open-prices/pull/1425).

`Price.category_tag` is only normalized to the taxonomy's canonical id once,
at creation time. When Open Food Facts renames a category's canonical id
(e.g. `en:kiwis` -> `en:kiwifruits`), old prices are never updated, so they
stop matching newer prices under the renamed id.

Adds a daily task that re-normalizes `category_tag` for all `type=category`
prices, following the existing `Model.update_task()` + `CRON_SCHEDULES`
convention (`Product`, `User`, `Location`):

- `Price.update_task()` / `Price.update_category_tag()` in `prices/models.py`
- `update_price_category_tag_task()` in `common/tasks.py`, daily at 02:40
- Tests in `PriceModelUpdateCategoryTagTest` (`prices/tests.py`)

### DIFF
--- a/open_prices/common/tasks.py
+++ b/open_prices/common/tasks.py
@@ -74,6 +74,10 @@ def update_location_counts_task():
     Location.update_task()
 
 
+def update_price_category_tag_task():
+    Price.update_task()
+
+
 def update_challenge_task():
     Challenge.update_task()
 
@@ -153,6 +157,7 @@ CRON_SCHEDULES = {
     "update_user_counts_task": ("0 2 * * *", {}),  # daily at 02:00
     "update_badge_task": ("5 2 * * *", {}),  # daily at 02:05
     "update_total_stats_task": ("10 2 * * *", {}),  # daily at 02:10
+    "update_price_category_tag_task": ("40 2 * * *", {}),  # daily at 02:40
     "update_location_counts_task": (
         "20 2 * * 1",  # every start of the week (at 02:20)
         {},
--- a/open_prices/prices/models.py
+++ b/open_prices/prices/models.py
@@ -16,6 +16,7 @@ from open_prices.challenges.models import Challenge
 from open_prices.common import (
     constants,
     history,
+    openfoodfacts,
     utils,
 )
 from open_prices.locations import constants as location_constants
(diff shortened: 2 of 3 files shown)


raphodn (approved):
awesome, thanks !

open_prices/common/tasks.py
```diff
@@ -153,6 +157,7 @@ def create_flags_from_price_outliers_and_update_view():
     "update_user_counts_task": ("0 2 * * *", {}),  # daily at 02:00
     "update_badge_task": ("5 2 * * *", {}),  # daily at 02:05
     "update_total_stats_task": ("10 2 * * *", {}),  # daily at 02:10
+    "update_price_category_tag_task": ("40 2 * * *", {}),  # daily at 02:40
```
raphodn, line 160: might change to a weekly basis after a few days, depending on the time it actually takes

PR #1438, opened 2026-09-15
chore(deps): remove some unused packages (pytest, pytest-cov, python-multipart...)

raphodn:
### What

Remove some unused packages : pytest, pytest-cov, blessed, croniter, python-multipart, httpx


TTalex:
@raphodn looks like croniter was still needed, tasks seem to fail according to sentry

raphodn:
> @raphodn looks like croniter was still needed, tasks seem to fail according to sentry

thanks !!! re-added in #1442

PR #1443, opened 2026-09-19
feat: import public data exports for local development

TaciteOFF:
The public data exports cannot be loaded with Django's `loaddata`. This adds an `import_public_data` command to load them into a local PostgreSQL database.

- Read locations, proofs and prices from JSONL or gzip files.
- Validate the files, then import them in one transaction. The locations, proofs and prices tables must be empty.
- Keep location, proof and price IDs and timestamps, link products by barcode, and rebuild counters and ID sequences.
- Use `--dry-run` to validate without writing data.
- Use `--allow-missing-references` for filtered exports. Missing links are set to `NULL`; without this option, they cause an error.
- Add tests and a guide with download and import commands.

Existing product details and local user accounts are kept. The command does not download product details or start background jobs.

Proof images are not downloaded or imported. Only their file paths are kept.

## Testing

- 36 Django tests passed on PostgreSQL 16.
- Ruff, scoped mypy checks and the strict documentation build passed.
- Imported fresh public exports locally: 7,280 locations, 123,319 proofs, 314,168 prices and 138,419 product references.
- Checked links, counters, timestamps and ID sequences. All 19 API checks passed, and category search displayed the imported prices in the frontend.


docs/guides/import-data.md
raphodn, line 1: ok so there's lots of text and code in 1 single PR ^^
I don't know if the verbosity should be in the docs, and/or in the code.

My intuition tells me that the docs should be lightweight-to-the-point, and direct towards the script that would hold the details of each option and how it works. As the script will be run by a few number of persons, and only devs, we can simplify the docs:
- show a few possible commands
- list a few one-liners of what it does and its choices
- move the rest of detailed doc to the script
- link to script

open_prices/common/import_data.py
raphodn, line 1: i would rename this file to `import_data`. to match with the rest of the changes (doc name, test & command names)

raphodn: which makes me think we could also have an `export_data` file and extract the logic of the `common/tasks` there

raphodn: and if we want them to sit together, then move them in a `common/data` folder ?
(this can be done in another PR later)

docs/guides/import-data.md
TaciteOFF: I refactored it

raphodn (approved):

open_prices/common/import_data.py
raphodn: did some light refactoring in #1446

